PR #15, opened 2015-07-26
README: missing bracket in `batwidget` example

JulianVolodia:
Added missing bracket and remove one whitespace (it would be more consistent)
Signed-off-by: Julian Volodia <julianvolodia@gmail.com>

---

Hi, I noticed when testing batt's stat wibox box ;)

BTW, why on the page (http://git.sysphere.org/vicious/about/) there is not mentioned about github repository? (and at all - "...you can send me..." - is ambiguous :-) )

Have a nice day! :)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,27 @@
 Full changelog is available online:
     http://git.sysphere.org/vicious/log/?showmsg=1
 ---------------------------------------------------
+6e16a41 bat: fix broken battery remaining time (was always N/A)
+bfbc1bd volume: get the normalized volume like alsamixer and DE's indicator
+3ef0f11 Next release, tag 2.1.2
+946271c bat: expose information on battery wear and tear
+7626989 README: write a list of major contributors to the project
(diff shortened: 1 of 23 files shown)


Mic92:
merged in https://github.com/Mic92/vicious/commit/240de45c4196f87bae3dc24f1a21e596eafabd80

This is not the official vicious github repo.
However I review patches send them via email upstream.
So I am like a proxy between anrxc and those who never used git-send-email before.

JulianVolodia:
> So I am like a proxy between anrxc and those who never used git-send-email before.
... so it's better to send patches via email only, yep?

Mic92:
You can if you want sent the patches directly to 
  Adrian C. (anrxc) <anrxc@sysphere.org>
However sometimes it might take some days before you get a response.

JulianVolodia:
I just asking for more preferred way by Devs for doing this and where the feedback could be given :) I was thinking the GH is better place to do it.

PR #18, opened 2016-08-04
Updated base URL for weather widget

khughitt:
Fixed base URL for weather widget: old URL no longer works and redirects to the tgftp.* URI.


muammar:
This has to be merged.  I was having problems with my widget showing `N/A` and after changing the URL everything works normal.

Mic92:
Thanks

PR #20, opened 2016-10-22
Added new wifi widget using iw

Mellich:
Sorry for the inconvenience, now the second try and only my additions:
On the Arch Linux Wiki pages the package 'wireless_tools' containing iwconfig is listed as deprecated.
https://wiki.archlinux.org/index.php/Wireless_network_configuration#Manual_setup
So I added a new widget to collect wifi information using iw instead.

### DIFF
--- a/widgets/wifiiw.lua
+++ b/widgets/wifiiw.lua
@@ -42,6 +42,7 @@ local function worker(format, warg)
         ["{rate}"] = 0,
         ["{freq}"] = 0,
 		["{txpw}"] = 0,
+		["{linp}"] = 0,
         ["{sign}"] = 0
     }
 
@@ -79,6 +80,8 @@ local function worker(format, warg)
       tonumber(string.match(iwresult, "freq: ([%d]+)") or winfo["{link}"])
     winfo["{sign}"] =  -- Signal level can be a negative value, don't display decibel notation
       tonumber(string.match(iwresult, "signal: (%-[%d]+)") or winfo["{sign}"])
+    winfo["{linp}"] =  -- Link Quality using the Windows definition (-50dBm->100%, -100dBm->0%)
+      100 - ((winfo["{sign}"] * -2) - 100)
     winfo["{txpw}"] =  -- TX Power can be a negative value, don't display decibel notation
       tonumber(string.match(iwresult, "txpower ([%-]?[%d]+)") or winfo["{txpw}"])
 
--- a/widgets/wifiiw.lua
+++ b/widgets/wifiiw.lua
@@ -1,0 +1,89 @@
+---------------------------------------------------
+-- Licensed under the GNU General Public License v2
+--  * (c) 2010, Adrian C. <anrxc@sysphere.org>
+---------------------------------------------------
+
+-- {{{ Grab environment
+local tonumber = tonumber
+local math = { ceil = math.ceil }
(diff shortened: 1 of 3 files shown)


widgets/wifiiw.lua
```diff
@@ -0,0 +1,89 @@
...
+    winfo["{ssid}"] =  -- SSID can have almost anything in it until new line
+      string.match(iwresult, 'SSID: ([^\n]*)') or winfo["{ssid}"]
+    winfo["{mode}"] =  -- everything after 'type ' until new line
+      string.match(iwresult, "type ([^\n]*)") or winfo["{mode}"]
+    winfo["{chan}"] =  -- Channels are plain digits
+      tonumber(string.match(iwresult, "channel ([%d]+)") or winfo["{chan}"])
+    winfo["{rate}"] =  -- We don't want to display Mb/s
+      tonumber(string.match(iwresult, "tx bitrate: ([%d%.]*)") or winfo["{rate}"])
+    winfo["{freq}"] =  -- Frequency are plain digits
+      tonumber(string.match(iwresult, "freq: ([%d]+)") or winfo["{link}"])
```
Mic92: `{link}` is never initialized here.

widgets/wifiiw.lua
```diff
@@ -0,0 +1,89 @@
...
+            if f then
+                iw = p .. "/iw"
+                f:close()
+                break
+            end
+        end
+    end
+
+    -- Get data from iw where available
+    local f = io.popen(iw .." dev ".. warg .. " link 2>&1;" .. iw .." dev ".. warg .. " info 2>&1")
```
Mic92: actually path search could be also delegated to the shell, which does also check if the file is executable.

```
io.popen("export PATH=$PATH:/sbin/:/usr/sbin:/usr/local/sbin; iw dev " ...
```

In that case the loop above, can be removed.

Also please escape shell characters from warg using `helpers.shellquote(warg)`.

widgets/wifiiw.lua
```diff
@@ -0,0 +1,89 @@
...
+local iwcpaths = { "/sbin", "/usr/sbin", "/usr/local/sbin", "/usr/bin" }
+-- }}}
+
+
+-- {{{ Wireless widget type
+local function worker(format, warg)
+    if not warg then return end
+
+    -- Default values
+    local winfo = {
```
Mic92: When I think about, we could also add an empty table here:

```
local winfo = {}
```

and then assign default values to fields like this:

```lua
winfo["{ssid}"] =  -- SSID can have almost anything in it until new line
    string.match(iwresult, 'SSID: ([^\n]*)') or "N/A"
```

widgets/wifiiw.lua
```diff
@@ -0,0 +1,89 @@
+---------------------------------------------------
+-- Licensed under the GNU General Public License v2
+--  * (c) 2010, Adrian C. <anrxc@sysphere.org>
```
Mic92: I don't think Adrian did this :)

widgets/wifiiw.lua
Mellich: actually there is no output of iw that equals the link quality provided by iwconfig. Or at least I didn't find it. Nevertheless, I added a {linp} field which calculates the link quality using the signal level.

Mic92:
Thanks!

PR #22, opened 2016-12-20
Add FreeBSD support

mutlusun:
Hello,

this is a first commit to discuss FreeBSD support further on a code basis. Following changes are introduced:
- Readme.md:
  - Headline of battery widget section is changed to a more neutral name (and I would like to order it alphabetically)
  - A list is added for the Linux and the FreeBSD widget to cover differences between them
  - Two helper functions and the FreeBSD widget are added

For me the following questions should be discussed:
- Is there a will to add FreeBSD specific code?
- It is really hard to determine the OS through lua. In my opinion it would be necessary that the user chooses the appropiate widget for him. To keep compatibility with the existing code I haven't renamed the Linux specific widget to `bat_linux`. So the question is how to distinguish properly between the same widgets for different OS (maybe OpenBSD may be added some time, or Linux distribution specific changes). For me it would be okay, to keep the old naming for Linux widgets and add to BSD specific widgets `_freebsd`. As an alternative, setting a global variable may be possible too, but this might not cover different features of widgets.

If there is any decision to merge this piece of code I would add the other FreeBSD widgets to this pull request to have it all in one place and normalize the README file accordingly.

Greetings!


README.md
```diff
@@ -146,6 +146,20 @@ argument given to vicious.register as the first argument, *warg* as
 the second, and return a table of values to insert in the format
 string.
 
+**Battery**
+
```
Mic92: I would suggest the following scheme:

- `vicious.widgets.bat`:
- supported platforms: linux, freebsd
  - provides state, charge, and remaining time for a requested battery
  - Arguments (per platform):
    - Linux: takes battery ID as an argument, i.e. "BAT0"
  - Returns:
    - returns 1st value as state of requested battery, 2nd as charge
      level in percent, 3rd as remaining (charging or discharging)
      time and 4th as the wear level in percent

helpers.lua
```diff
@@ -12,7 +12,10 @@ local pairs = pairs
 local rawget = rawget
 local require = require
 local tonumber = tonumber
-local io = { open = io.open }
+local io = { 
+    open = io.open, 
+    popen = io.popen 
+}
```
Mic92: here `uname` is used to distinguish the operating system once. This information should be also exposed to the user in case configuration should work on multiple platforms. We also need a helper, which set a platform specific function at the end of the module ```return setmetatable(fs, { __call = function(_, ...) return worker(...) end })``
Platform specific code should be in dedicated files suffixed by the platform for example `_linux`, `_freebsd`,
the helper gets the supported platforms as arguments and will either load such files or return an error (platform not supported).

mutlusun:
Hey!

You might wanna have a look on these changes. I changed the readme structure and added code to load platform specific code (`helpers.wrequire()`). I added some more moduls to test the changes (right now only bat, fs, date and volume widget are usable).

But I run into errors I could not fix, only the date widget is affected. If I try to use the date widget, awesome crashes with an error `attempt to call field date (a nil value)` in line 23. `os.date` is indeed a nil value. But I don't know why, because I haven't changed there anything. Maybe it has something with the naming of the widgets/modules (if I try to convert the os widget I run into the same problems)? Do you have any clue here?

What do you think about the code to load platform specific code? There are no changes needed, except of renaming the widgets and it is easily extendable. And I was unsure about the best way to raise errors in this module, I used the assert function now (I was thinking about gears.debug from awesome but that would add dependencies -- even gears.timer is used in init.lua).

Best greetings!

helpers.lua
```diff
@@ -35,10 +37,44 @@ local scroller = {}
...
+-- }}}
+
 -- {{{ Loader of vicious modules
 function helpers.wrequire(table, key)
-    local module = rawget(table, key)
-    return module or require(table._NAME .. "." .. key)
+    local ret = rawget(table, key)
+
+    if not ret then
+        os = helpers.getos()
```
Mic92: we probably need to call `uname` once on startup and use the cached value

Mic92:
we probably also should get rid off the old README. I have not looked in depth into the made changes, but I agree so far on the structure. I will also have a look at the os.date problem.

helpers.lua
mutlusun: If I got it right, this code should only be called once for every widget (at creation time), so I think it isn't such an overhead. But using a cached value is an option, too!

The other commits change this function a bit to solve the problem more elegant.

mutlusun:
Hello Mic92,

now, I ported almost all widgets (some are still missing), but I investigated further into the bug that prevented the date widget to be usable. A simple
```Lua
local os = require("os")
```
in the `date_all.lua` file worked around this issue. Do you have any explanation for this? Do you mind if I add this lines to all widgets where it is needed?

Best

Mic92:
@mutlusun no, I cannot it explain it. But requiring it explicit does not hurt, it is even a little faster because the accessing a local scope is faster when doing a method lookup. I will take a short at your code now.

mutlusun:
Okay, I think we're done! Please write if you find any bugs or problems ...

The os bug is fixed, strange enough there were only problems with os not with io or string, etc.

widgets/net_freebsd.lua
```diff
@@ -0,0 +1,85 @@
...
+
+        local split = { line:match(("([^%s]*)%s*"):rep(12)) }
+
+        if buffer == nil then
+            buffer = { tonumber(split[8]), tonumber(split[11]) } -- recv (field 8) and send (field 11)
+        else
+            buffer = { buffer[1] + tonumber(split[8]), buffer[2] + tonumber(split[11]) }
+        end
+
+        ::continue::
```
Mic92: this would break lua 5.1, for the moment I would stick to:

```
for line in f:lines() do
      if not (line:find("<Link") and line:find("Name ")) then -- skipping missleading lines
          # ...
      end
end 
```

widgets/thermal_freebsd.lua
```diff
@@ -13,14 +13,19 @@ local thermal_freebsd = {}
 -- {{{ Thermal widget type
 local function worker(format, warg)
     if not warg then return end
+    local thermals = {}
+    local cnt = 1
+    while cnt <= #warg do
+        local output = helpers.sysctl( "" .. warg[cnt] .. "" )
```
Mic92: what is the purpose of the string interpolation here?

widgets/bat_acpi_freebsd.lua
```diff
@@ -3,12 +3,15 @@
...
 local bat_acpi = {}
 
 local function worker(format)
+    local battery = "batt"
+    if warg then
+        battery = warg
+    end
     local bat_info = {}
-    local pcall = "acpiconf -i batt"
+    local pcall = "acpiconf -i " .. battery
```
Mic92: battery argument should be escaped. Take a look at `helpers.shellquote`

widgets/bat_freebsd.lua
```diff
@@ -1,48 +1,65 @@
...
-
-
--- {{{ Battery widget type
 local function worker(format)
+    local battery = "batt"
+    if warg then
+        battery = warg
+    end
+    local bat_info = {}
+    local pcall = "acpiconf -i " .. battery
```
Mic92: escaping like mentioned before.

mutlusun:
Er, I'm sorry about the continue statement ... I simply forgot to remove it. I did correct escaping now. Be aware that some later commits change the reviewed code (for example the string interpolation).

widgets/thermal_freebsd.lua
mutlusun: this is already fixed in a later commit

Mic92:
Are the external modules like io no imported correctly in all widgets?

mutlusun:
Yes, they are all working now!

Mic92:
I have rebased your commits for easier code review (except for the README, which was not as important and I just introduce all the changes in one commit at the end)

```
https://github.com/Mic92/vicious/commits/freebsd_port
```

If you agree on the result, I would merge this.

Mic92:
Mhm. Now I have the same bug regarding in my own `rc.lua`. This should be fixed or we break other users configuration.

Mic92:
The bug is fixed. In your code you overwrote the global `os` instead of using a local variable. I also ported all contrib widgets to the new scheme.

Mic92:
superseeded by #27

PR #23, opened 2017-01-12
Update README

walken:
CPU example has an error


Mic92:
I think we should drop "README", I inherit it from the origin project.

walken:
Oh yes, right !

Mic92:
now it is gone.

PR #28, opened 2017-02-15
fixing `nil` sent to curl

starenka:


Mic92:
I wonder, how this has ever worked. Thank you, for your patches

starenka:
No problem, cheers!

PR #29, opened 2017-02-17
Handle single widget arguments to force

supplantr:


mutlusun:
Hey,

thanks for your contribution! Makes sense as cache and other functions take only a single argument ...

supplantr:
Oops, this doesn't actually work. Any single argument will be a wibox, so still of type table...
I think the only way to robustly handle this use case is to make `force` take varargs, and unpack a table of multiple widgets to achieve what's currently possible.

Mic92:
I have reverted this pull request for the moment.

mutlusun:
Urgh, I'm sorry about merging too fast, thought it's only a little change ...

supplantr:
Yeah, sorry about the noise. Should I open a new issue to resolve handling of single arguments, or just not worry about it? The current API is fine, but I've never needed to force multiple widgets at once.

Mic92:
I guess the current API is good enough. If I would implement it from scratch, I probably would implement it with a single argument, but changing it now does not justify breaking other peoples code.
Also I don't see how varargs would solve this `abc("foo", "bar")` is not the same as `abc({"foo", "bar"})`.

PR #30, opened 2017-02-23
get custom data from a widget

mutlusun:
Hello Mic92,

I added a small function to get a custom string from a widget, that allows following in awesome/rc.lua:
```lua
mybattery = wibox.widget.textbox()
vicious.register(mybattery, vicious.widgets.bat, "<span color=\"#ffc48c\">$2%$1</span> ($5W)", 17, "0")
mybattery:buttons(awful.util.table.join(
    awful.button({ }, 1, function()
        naughty.notify({ preset = naughty.config.presets.normal,
                         title = "Battery indicator",
                         text = vicious.getdata(vicious.widgets.bat, "Remaining time: $3h\nWear level: $4%\nPresent rate: $5W", "0") })
    end)
))
```

What do you think about these changes? If you agree, I will add a documentation for it in the readme file. I'm not really happy with the function name (`getdata()`), do you know something better?

In the future I will keep it this way: I will commit directly small fixes in the freebsd files but all greater changes or API changes I will introduce through pull requests. So you have a chance to look over it ...

Best
mutlusun

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -261,11 +261,15 @@ end
 -- }}}
 
 -- {{{ Get custom widget format data
-function vicious.getdata(myw, format, warg)
-    return helpers.format(format, myw(format, warg))
+function vicious.call(myw, format, warg)
+    local mydata = myw(format, warg)
+    if type(format) == "string" then
+        return helpers.format(format, mydata)
+    elseif type(format) == "function" then
+        return format(myw, mydata)
+    end
 end
 -- }}}
-
 
 return vicious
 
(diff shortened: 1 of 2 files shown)


init.lua
```diff
@@ -260,6 +260,13 @@ function vicious.activate(widget)
 end
 -- }}}
 
+-- {{{ Get custom widget format data
+function vicious.getdata(myw, format, warg)
```
Mic92: I would name it `vicious.call`.

Mic92:
Thanks!

PR #31, opened 2017-02-25
contrib: make pulse round rather than floor volume

jmccrohan:
Check volpercent to see if ceil or floor should be used.

Signed-off-by: Jonathan McCrohan <jmccrohan@gmail.com>


Mic92:
Thanks!

PR #32, opened 2017-03-30
Minimal change to allow asynchronous call (for Awesome4+)

getzze:
If a `[widget].async` function is defined, it is called instead of the synchronous `worker` function. It works only for awesome>4.0 as it uses `awful.spawn.easy_asunc`. It takes two arguments, `warg` and a `callback` function that updates the widget.

Only applied with one example, `pkg`.

I also added another value to the table output of `pkg`, the list of packages.

By the way, the first argument `format` of the `worker` functions is never used, except for the `date` widget. Maybe it would be good to remove it to simplify the code.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -75,6 +75,8 @@ local function update(widget, reg, disablecache)
         if t and cache then
             cache.time, cache.data = t, data
         end
+        -- Unlock task
+        reg.lock = false
     end
     
     -- Check for cached output newer than the last update
@@ -83,7 +85,10 @@ local function update(widget, reg, disablecache)
         return update_value(widget, format_data(reg.format, c.data))
     elseif reg.wtype then
         if reg.wtype.async then
-            return reg.wtype.async(reg.warg, function (data) update_value(format_data(reg.format, data), widget, t, c) end)
+            if not reg.lock then
+                reg.lock = true
+                return reg.wtype.async(reg.warg, function (data) update_value(format_data(reg.format, data), widget, t, c) end)
+            end
         else
             return update_value(format_data(reg.format, reg.wtype(nil, reg.warg)), widget, t, c)
         end
@@ -154,6 +159,7 @@ function vicious.register(widget, wtype, format, timer, warg)
     local reg = {
         -- Set properties
         wtype  = wtype,
+        lock   = false,
         format = format,
         timer  = timer,
         warg   = warg,
--- a/init.lua
+++ b/init.lua
@@ -49,19 +49,19 @@ local function update(widget, reg, disablecache)
     local t = os.time()
     local data = {}
 
-    local function format_data(rformat, data)
+    local function format_data(data)
         local ret
         if type(data) == "table" then
-            if type(rformat) == "string" then
-                ret = helpers.format(rformat, data)
-            elseif type(rformat) == "function" then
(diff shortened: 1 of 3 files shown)


init.lua
```diff
@@ -49,38 +49,45 @@ local function update(widget, reg, disablecache)
...
-    else
-        widget.text = data
+    
+    -- Check for cached output newer than the last update
+    local c = widget_cache[reg.wtype]
+    if c and c.time and c.data and t < c.time+reg.timer and not disablecache then
+        return update_value(widget, format_data(reg.format, c.data))
+    elseif reg.wtype then
+        if reg.wtype.async then
+            return reg.wtype.async(reg.warg, function (data) update_value(format_data(reg.format, data), widget, t, c) end)
```
Mic92: async should be not called again, when the same widget function is still running.
Maybe you can set a flag to avoid that, which is unset after execution.

widgets/pkg_all.lua
```diff
@@ -36,12 +37,53 @@ local function worker(format, warg)
...
+    size = math.max(size-first, 0)
     f:close()
 
-    return {_pkg.sub and math.max(updates-_pkg.sub, 0) or updates}
+    return {size, lines}
+end
+-- }}}
+
+-- {{{ Packages widget type
+function pkg_all.async(warg, callback)
```
getzze: I added a synchronous variant that uses the async code

init.lua
getzze: I added a `lock` variable in the `reg` table.
Also there were errors with the arguments of `update_value`, it should be corrected now.

widgets/pkg_all.lua
```diff
@@ -33,16 +33,34 @@ local function worker(format, warg)
...
-    end
-    f:close()
+-- {{{ Packages widget type
+local function worker(format, warg)
+    local ret = nil
+    
+    pkg_all.async(warg, function(data) ret = data end)
 
-    return {_pkg.sub and math.max(updates-_pkg.sub, 0) or updates}
+    while ret==nil do end
```
mutlusun: Would it be better to give the `async` function an callback function from `init.lua`? So it could update the widget itself and we would not need to waste CPU cycles ... (don't know how handy and implementable this is)

mutlusun:
Hey,

thanks for your work and your contribution! I added two little notes (there are only suggestions). I agree with you, that we should remove the format argument.

Should we make a release before merging this? So everyone who has to use the old awesome version can use it. If not I would add a notice to the readme file which commit is the last working with awesome 3.5.

widgets/pkg_all.lua
getzze: This is just to maintain the old behaviour of the `worker` function, but it is never actually called.
I think the best would be to have only the function async for all the widgets (or change the `worker` functions to take a callback as argument). Like that all the widgets are asynchronous. It should not be too complicated to implement: with a function `parse` when io.popen was used, or calling directly `callback(...)` at the end if all the code is synchronous.
The only problem is that the API would change, so all the widgets have to be rewritten...

Mic92:
We already had v2.2.0 recently. This will be the last version for awesome 3.5.

Mic92:
Thanks for your contribution!

PR #33, opened 2017-04-01
Use wibox for battery widget example

kressi:
Since awesome 4 example batterywidget in an awful.widget.progressbar does not work anymore. (left progressbar in below screenshot) Instead, batterywidget has been added with wibox.widget.progressbar and is then put into an wibox.widget. (progressbar on the right in screenshot) 

That is how I got battery widget to work again. Correct me if there is a better way to do it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -628,17 +628,27 @@ argument
 
 **Battery widget**
 
-```Lua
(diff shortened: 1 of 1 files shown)


Mic92:
Thanks

PR #40, opened 2017-07-17
Remove format argument

mutlusun:
Hello,
as discussed in #32 it would be nice to remove the format argument as it is most of the time not needed. This is done in this pull request.

The only really change had to be done in date widget. What do you think about these changes? Is there a better way to do this? The change in the date widget breaks the API, it need to be called in `rc.lua` now like that (see also explanation in readme):
```lua
mydate = wibox.widget.textbox()
vicious.register(mydate, vicious.widgets.date, "<span color=\"#ffffff\">$1</span>", 61, { "%b %d, %R" })
```
What about merging this into a new branch and prepare a v4-only version? We could remove the old helper functions and move to the new async function.

Best

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,7 +588,7 @@ order to actually display them.
 **Date widget**
 
 ```Lua
-    datewidget = widget({ type = "textbox" })
+    datewidget = wibox.widget.textbox()
     vicious.register(datewidget, vicious.widgets.date, "%b %d, %R")
 ```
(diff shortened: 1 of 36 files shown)


Mic92:
Some people seem to use that for the date widget. So I added 1cfa040fbd2b1740e02357f54a10f83ea36af0b1 instead.

PR #43, opened 2017-10-05
Fixed function indexing

empijei:
wtype is a function and if indexed without parens to call it, crashes awesome.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -82,7 +82,7 @@ local function update(widget, reg, disablecache)
     if c and c.time and c.data and t < c.time+reg.timer and not disablecache then
         return update_value(format_data(c.data))
     elseif reg.wtype then
-        if reg.wtype.async then
+        if reg.wtype().async then
             if not reg.lock then
                 reg.lock = true
                 return reg.wtype.async(reg.warg, 


empijei:
image
This is what happens if I load vicious without those parens, which does not happen if I call the function instead.

This is my [rc.lua](https://github.com/empijei/dotfiles/blob/master/awesome/rc.lua) if you need it, which crashes at [this line](https://github.com/empijei/dotfiles/blob/master/awesome/rc.lua#L298)

init.lua
```diff
@@ -82,7 +82,7 @@ local function update(widget, reg, disablecache)
     if c and c.time and c.data and t < c.time+reg.timer and not disablecache then
         return update_value(format_data(c.data))
     elseif reg.wtype then
-        if reg.wtype.async then
+        if reg.wtype().async then
```
Mic92: This will run the function, which is not what we want. In our case widgets are tables, where the `__call` attribute is set so they are also callable.

Mic92:
I added a comment here: https://github.com/empijei/dotfiles/commit/5b2e0692395fab80db4680e65d3580c7655b60b7#commitcomment-24805719

empijei:
Thanks, now I get how it is supposed to work. Sorry for the misunderstanding.

empijei:
I changed my code following your suggestions and most of it works, just do you have any idea why [this](https://github.com/empijei/dotfiles/blob/master/awesome/rc.lua#L434) is not using the format string anymore?

I'm only getting this output with any string I provide as format.
image

Issue #41, opened 2017-07-23
Volume Widget shellquote issue


bhartshorn:
When I pass "Master -D pulse" as wargs to vicious.widgets.volume, the helpers.shellquote function makes the final amixer command `amixer -M get 'Master -D pulse'`. This command line just returns 
`amixer: Unable to find simple control 'Master '-D pulse'',0`

I noticed in the commit history that the shellquote helper function was added to the volume widget in 2014. Am I missing something obvious about how to configure this widget, or is this a long-standing bug?

Mic92:
At the moment we only support adding a single argument to this widget.

PR #44, opened 2017-10-08
Add device configuration option to volume widget

bhartshorn:
Hi. This is in reference to my issue #41. 

I'm a total noob at Lua, but I think I ironed out any bugs I introduced. Specifically I was confused on using _local_, so I'm not sure if I came up with a good solution. Point me the right way and I will fix it. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,9 +448,9 @@ Supported platforms: Linux (required tool: amixer), FreeBSD.
 
 - Arguments (per platform):
   * Linux: takes either a single argument containing the ALSA mixer control as
-    an argument, i.e. `"Master"`, or a table, with the first argument being the
-    ALSA mixer control and the second being the ALSA device ID, i.e.
-    `{"Master", "pulse"}`.
+    an argument, i.e. `"Master"`, or a table passed as command line arguments
+    to [amixer(1)](https://linux.die.net/man/1/amixer),
+    i.e `{"PCM", "-c", "0"}` or `{"Master", "-D", "pulse"}`.
   * FreeBSD: takes the mixer control as an argument, i.e. `"vol"`
 - Returns:
   * Linux: returns 1st value as the volume level and 2nd as the mute state of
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -447,8 +447,10 @@ Provides volume levels and state of requested mixers.
 Supported platforms: Linux (required tool: amixer), FreeBSD.
 
 - Arguments (per platform):
-  * Linux: takes the ALSA mixer control as an argument, i.e. `"Master"`,
-    optionally append the card ID or other options, i.e. `"PCM -c 0"`
+  * Linux: takes either a single argument containing the ALSA mixer control as
+    an argument, i.e. `"Master"`, or a table, with the first argument being the
+    ALSA mixer control and the second being the ALSA device ID, i.e.
+    `{"Master", "pulse"}`.
```
Mic92: Btw. we have also a dedicated pulseaudio widget:
https://github.com/Mic92/vicious/blob/master/contrib/pulse_all.lua

README.md
```diff
@@ -447,8 +447,10 @@ Provides volume levels and state of requested mixers.
 Supported platforms: Linux (required tool: amixer), FreeBSD.
 
 - Arguments (per platform):
-  * Linux: takes the ALSA mixer control as an argument, i.e. `"Master"`,
-    optionally append the card ID or other options, i.e. `"PCM -c 0"`
+  * Linux: takes either a single argument containing the ALSA mixer control as
+    an argument, i.e. `"Master"`, or a table passed as command line arguments
+    to [amixer(1)](https://linux.die.net/man/1/amixer),
+    i.e `{"PCM", "-c", "0"}` or `{"Master", "-D", "pulse"}`.
```
Mic92: How about the following: this makes it more flexible.

widgets/volume_linux.lua
```diff
@@ -26,8 +26,22 @@ local function worker(format, warg)
         ["off"] = "♩"  -- "M"
     }
 
+    local mixer = nil
+    local device = nil
```
Mic92: using local is always a good idea. The default value of a variable should be nil already.

README.md
bhartshorn: Looks good and the command seems to work fine with the quotes on my end. I like this method better. Thanks.

PR #45, opened 2017-11-16
Readme file update.

BeniaminK:
There is a dead link in the README file and in fact, the require("vicious.widgets") in the "lua" doesn't work, while require("vicious.widgets.init") does


Mic92:
Thanks!

PR #46, opened 2017-11-20
Add widget for cmus (follow mpd conventions)

JuanKman94:
I wrote this platform independent widget for `cmus`, mostly based on the `mpd` widget.
Also fixed some markup issues with the contrib/README in the same commit

Tested on Fedora 26
```
awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.4 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.2
```

### DIFF
--- a/contrib/cmus_all.lua
+++ b/contrib/cmus_all.lua
@@ -31,8 +31,13 @@ local function worker(format, warg)
     }
 
     -- Fallback to CMUS defaults
-    local host = warg and (warg.host or warg[1]) or
-		os.getenv("XDG_RUNTIME_DIR") .. "/cmus-socket"
+    local host = warg and (warg.host or warg[1]) or os.getenv("CMUS_SOCKET")
+
+    if not host and os.getenv("XDG_RUNTIME_DIR") then
+        host = os.getenv("XDG_RUNTIME_DIR") .. "/cmus-socket"
+    elseif not host then
+        host = os.getenv("HOME") .. "/.config/cmus/socket"
+    end
 
     -- Get data from CMUS server
     local f = io.popen(string.format("cmus-remote --query --server %s", host))
--- a/contrib/cmus_all.lua
+++ b/contrib/cmus_all.lua
@@ -1,0 +1,69 @@
+-----------------------------------------------------------
+-- Licensed under the GNU General Public License v2
+--  * (c) 2017, JuanKman94 <juan.carlos.menonita@gmail.com>
+-----------------------------------------------------------
+
+-- {{{ Grab environment
+local tonumber = tonumber
+local io = { popen = io.popen }
+local setmetatable = setmetatable
+local string = { gmatch = string.gmatch, format = string.format }
+local helpers = require("vicious.helpers")
+-- }}}
+
+
+-- Cmus: provides CMUS information
+-- vicious.widgets.cmus
+local cmus_all = {}
+
+
+-- {{{ CMUS widget type
+local function worker(format, warg)
+    local cmus_state  = {
+        ["{duration}"]  = 0,
+        ["{file}"] = "N/A",
+        ["{status}"] = "N/A",
+        ["{title}"]  = "N/A",
+        ["{artist}"]  = "N/A",
+        ["{continue}"]  = false,
+        ["{shuffle}"]  = false,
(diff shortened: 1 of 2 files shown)


contrib/cmus_all.lua
```diff
@@ -0,0 +1,69 @@
...
+        ["{shuffle}"]  = false,
+        ["{repeat}"]  = false,
+    }
+
+    -- Fallback to CMUS defaults
+    local host = warg and (warg.host or warg[1]) or
+		os.getenv("XDG_RUNTIME_DIR") .. "/cmus-socket"
+
+    -- Get data from CMUS server
+    local f = io.popen(string.format("cmus-remote --query --server %s", host))
```
Mic92: Can this block? If yes, there is also `spawn.easy_async` see `pkg_all.lua` for an example.

contrib/cmus_all.lua
```diff
@@ -0,0 +1,69 @@
...
+        ["{title}"]  = "N/A",
+        ["{artist}"]  = "N/A",
+        ["{continue}"]  = false,
+        ["{shuffle}"]  = false,
+        ["{repeat}"]  = false,
+    }
+
+    -- Fallback to CMUS defaults
+    local host = warg and (warg.host or warg[1]) or
+		os.getenv("XDG_RUNTIME_DIR") .. "/cmus-socket"
```
Mic92: Where will cmus write to if `XDG_RUNTIME_DIR` is not set?

contrib/cmus_all.lua
JuanKman94: It cannot block, if the server is not reached it returns immediately

contrib/cmus_all.lua
JuanKman94: Good observation, tested in a non-XDG environment and it crashed. I'll make a commit

Mic92:
I added here two commits to prevent some crashes: https://github.com/Mic92/vicious/commits/cmus

contrib/cmus_all.lua
```diff
@@ -0,0 +1,74 @@
...
+                cmus_state["{"..module.."}"] = helpers.escape(value)
+            elseif module == "duration" then
+                cmus_state["{"..module.."}"] = tonumber(value)
+            elseif module == "status" then
+                cmus_state["{"..module.."}"] = value
+            else
+                for k, v in string.gmatch(value, "([%w]+) (.*)$") do
+                    if module == "tag" then
+                        if k == "title" or k == "artist" then
+							cmus_state["{"..k.."}"] = helpers.escape(v)
```
Mic92: I am not sure about escaping here. I see `&quot;` instead of `"` in vicious.

JuanKman94: Weird, I suppose the song title has quotes. Then perhaps we could let the format function handle the escaping

Mic92: Probably soon, I will upgrade to newer awesome v4 API and have to break with awesome 3 anyway.

Mic92: We cannot do escaping in the format function, since user might want to add markup.

JuanKman94: Oh, I meant the user's format function, not the internal one. I'll go ahead and remove the escape

contrib/cmus_all.lua
```diff
@@ -0,0 +1,78 @@
...
+    if not host then
+      if os.getenv("XDG_RUNTIME_DIR") then
+        host = os.getenv("XDG_RUNTIME_DIR") .. "/cmus-socket"
+      else
+        host = os.getenv("HOME") .. "/.config/cmus/socket"
+      end
+    end
+
+    -- Get data from CMUS server
+    local f = io.popen("cmus-remote --query --server " .. helpers.escape(host))
```
Mic92: I took the wrong function, this should have been `helpers.shellquote`.

Mic92:
94f91c57ab83701c038d8ae0499b0d89996e26f1

contrib/cmus_all.lua
JuanKman94: Sorry I couldn't see this earlier

PR #47, opened 2017-11-22
Bitcoin price widget

0x5b:
Simple text widget. Provides a price in any currency by code:
(https://en.wikipedia.org/wiki/ISO_4217)

Maybe someone needs.
Did not test for freebsd...


widgets/btc_linux.lua
```diff
@@ -0,0 +1,54 @@
...
+
+    -- Default values
+    local btc = { ["{price}"] = "N/A" }
+    local response, code = nil, nil
+
+    local currency_code = string.upper(warg)
+    local url = "https://api.coindesk.com/v1/bpi/currentprice/"..helpers.escape(currency_code)..".json"
+
+    -- Get current price
+    response, code = https.request(url)
```
Mic92: Wow do you avoid blocking here, when network is not reachable?

0x5b: I dont know lua well, but testing in lua repl showed: if network is not reachable, ```https.request()``` returns ```nil``` instantly. I could be wrong...

0x5b: well, in weather widget, information was received by ```curl```, i can do the same

Mic92: the weather widget is also buggy in this regard as it will also block. You can look at `pkg_all.lua` to see how commands can be called asynchronous.

0x5b: @Mic92 , done, looks like async call..

widgets/btc_linux.lua
```diff
@@ -0,0 +1,57 @@
...
+
+-- {{ Bitcoin widget type
+function btc_linux.async(format, warg, callback)
+    -- Default values
+    if not warg then warg = "usd" end
+
+    local btc = { ["{price}"] = "N/A" }
+    local currency_code = string.upper(warg)
+    local url = "https://api.coindesk.com/v1/bpi/currentprice/"..helpers.escape(currency_code)..".json"
+    local cmd = "curl --connect-timeout 1 "..url
```
Mic92: the function name `escape` here is missleading and not the right function for shell escaping.
Use shellquote instead:

```diff
-    local url = "https://api.coindesk.com/v1/bpi/currentprice/"..helpers.escape(currency_code)..".json"
-    local cmd = "curl --connect-timeout 1 "..url
+    local url = "https://api.coindesk.com/v1/bpi/currentprice/" .. currency_code .. ".json"
+    local cmd = "curl "..helpers.shellquote(url)
```

`connect-timeout` is also not necessary, when the call is asynchronous.

widgets/btc_linux.lua
```diff
@@ -0,0 +1,57 @@
+---------------------------------------------------
+-- Licensed under the GNU General Public License v2
+--  * (c) 2017, 0x5b <dragen15051@gmail.com>
+---------------------------------------------------
+
+-- {{{ Grab environment
+local setmetatable = setmetatable
+local pcall = pcall
+local helpers = require("vicious.helpers")
+local https = require("socket.http")
```
Mic92: This one can be removed then.

widgets/btc_linux.lua
```diff
@@ -0,0 +1,57 @@
...
+        return btc
+    end
+    -- }}
+
+    spawn.easy_async(cmd, function(stdout) callback(parse(stdout)) end)
+end
+-- }}}
+
+-- {{ Bitcoin widget type
+local function worker(format, warg)
```
Mic92: The synchronous variant would look a bit different, see `widgets/pkg_all.lua`. I think we do not need this function at all. `worker` and `setmetatable` can be removed here.

Just add `return worker` as the last statement.

0x5b: I dont understand it, nothing works without ```setmetatable```

contrib/btc_all.lua
```diff
@@ -0,0 +1,51 @@
...
+        btc["{price}"] = string.sub(data["bpi"][currency_code]["rate"], 0, -3)
+        return btc
+    end
+    -- }}
+
+    spawn.easy_async(cmd, function(stdout) callback(parse(stdout)) end)
+end
+-- }}}
+
+return btc_all
```
Mic92: I meant to say, that you only need to return the module here. The metatable is not required for the asynchronous variant.

contrib/README.md
```diff
@@ -58,6 +58,18 @@ Supported platforms: Linux (required tools: `sysfs`)
 
 **vicious.contrib.batproc**
 
+**vicious.contrib.btc**
+
+Provides current Bitcoin price in any currency by code (
+https://en.wikipedia.org/wiki/ISO_4217).
+Supported platforms: Linux, FreeBSD.
```
Mic92: it should mention `curl` and your lua library as dependencies.

0x5b: ok, i updated required and squash commits

contrib/btc_all.lua
0x5b: oh, tnx... sorry, its my first try to code in lua

contrib/README.md
```diff
@@ -60,6 +60,21 @@ Supported platforms: Linux (required tools: `sysfs`)
 
 **vicious.contrib.batproc**
 
+**vicious.contrib.btc**
+
+Provides current Bitcoin price in any currency by code (
+https://en.wikipedia.org/wiki/ISO_4217).
+Requires `curl` and one of the following json libraries:
+ - [lua-cjson](https://github.com/mpx/lua-cjson/))
+ - [luajson](https://github.com/harningt/luajson/)
```
Mic92: `luajson` was not packaged yet for me. So I tested with `cjson` instead.

PR #52, opened 2018-01-20
Data caching fixes

BeniaminK:
Problem fixed by the pull request:
Vicious cache held values returned from format_data() instead of the data itself, so when many widgets were using the same wtype with other `format` string or function, they didn't work correctly (returning only first cached formatted string)

Changes include:
* caching raw worker data first, then updating values for the widgets
* Small readability changes in the vicious.register
* initial data set for timer (so c.time and c.data is always available and is updated since the first update due to initial time=0 value)

Backward compatibility preserved.


Mic92:
Thanks!

Issue #54, opened 2018-03-08
FreeBSD support for memory widget w/patch


dkrand:
I couldn't get the mem.lua widget to easily work well on FreeBSD with /compat/linux/proc/meminfo via the linprocfs filesystem. The memory stats seemed to be wildly wrong. So I leveraged your sysctl helper and hacked together a mem_freebsd.lua module. It doesn't support buffer memory (since I didn't find source of that data) but otherwise does everything else that mem.lua does.

I couldn't find a nice and easy way to get summary swap information, but meminfo via linprocfs does that, so I just used that.

dkrand:
[mem_freebsd.lua.txt](https://github.com/Mic92/vicious/files/1793765/mem_freebsd.lua.txt)

Mic92:
cc @mutlusun

mutlusun:
Hello,
I'm sorry for the late reply. As @psycorama pointed out, the current implementation for FreeBSD uses sysctl as provider. So there is no need to use the compat layer. But I tested this widget again and I found several errors in the reported numbers. For example the first argument doesn't give the used memory, it gives the free memory. I will provide a pull request.

psycorama:
Hy again.

I do agree with @mutlusun pointing out bugs, but i have to disagree with his assessment for the first returned argument. 
I'm a bit under the weather at the moment, but i will provide another pull request within the next days.

So, maybe hold back with merging for a bit?

dkrand:
On 03/09/18 05:57, Andreas Geisenhainer wrote:
> What's wrong with the current implementation for FreeBSD?

Blush!  Mainly the fact that I didn't look at the most recent version of 
Viscious.   Sorry about that. I had 2.1.3, not the master branch. Never even 
looked at the master branch I'm embarrassed to say.

But, there are a few problems with it though.  Swap space usage isn't 
calculated correctly. It uses vm.swapin out of the vm.stats table, and the 
description of that OID is:

% sysctl -d vm.stats.vm.v_swapin
vm.stats.vm.v_swapin: Swap pager pageins

Which is not the amount of swap space in use. I could not find a sysctl that 
reported swap space in use. There is the system call kvm_swapinfo(3) and the 
swapinfo(8) command both of which seem to do the right thing, but I didn't 
want to figure out how to call kvm_swapinfo(3) from Lua. Nor did I want the 
hassle of having to parsing the output of swapinfo and summing each of the 
rows. (With multiple swap devices, which admittedly is rare these days, 
[...]

PR #55, opened 2018-03-10
freebsd memory clean ups and correct calculation

mutlusun:
This pull request will correct the memory calculation of the freebsd memory widget and cleans up code. There is small change in the battery widget as well (unconsidered state).

I would be happy if someone could test it!

See issue #54 for further discussions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,9 +323,8 @@ Supported platforms: Linux, FreeBSD.
     total system memory, 4th as free memory, 5th as swap usage in percent, 6th
     as swap usage, 7th as total system swap, 8th as free swap and 9th as
     memory usage with buffers and cache
-  * FreeBSD: see above, but 10th value as memory usage with buffers and cache
-    as percent and 11th value as wired memory (memory used by kernel) is
-    reported
+  * FreeBSD: see above, but 9th value is not returned (always `-1`) and there
+    is a 10th value giving wired memory
 
 **vicious.widgets.mpd**
 
--- a/widgets/bat_freebsd.lua
+++ b/widgets/bat_freebsd.lua
@@ -30,6 +30,8 @@ local function worker(format, warg)
         state =  "↯"
     elseif bat_info["State"] == "charging" then
         state = "+"
+    elseif bat_info["State"] == "critical charging" then
+        state = "+"
     elseif bat_info["State"] == "discharging" then
         state = "-"
     else
--- a/widgets/mem_freebsd.lua
+++ b/widgets/mem_freebsd.lua
@@ -16,6 +16,7 @@ local function worker(format)
     local vm_stats = helpers.sysctl_table("vm.stats.vm")
     local _mem = { buf = {}, total = nil }
 
+    -- Get memory space in bytes
     _mem.total = tonumber(vm_stats.v_page_count) * pagesize
     _mem.buf.f = tonumber(vm_stats.v_free_count) * pagesize
     _mem.buf.a = tonumber(vm_stats.v_active_count) * pagesize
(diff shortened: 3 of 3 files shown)


Mic92:
cc @dkrand

psycorama:
I've tested it, and I suggest to hold back with merging.

Having taken another look at https://wiki.freebsd.org/Memory and the local outputs of, I would like to 
take another run at it.

For the time being: 
* should we discuss what should count as Used memory?
* where to put Laundry?
* and how to handle Wired memory?

mutlusun:
Hello,
thanks for your thoughts about that topic! As the calculation was wrong over a year now, there is now need to hurry ;)

I used [this link](https://www.cyberciti.biz/faq/freebsd-command-to-get-ram-information/) as reference but in doubt I would more trust the FreeBSD wiki. In their [script](https://www.cyberciti.biz/files/scripts/freebsd-memory.pl.txt) they calculated free memory as in the proposed patch. This calculation correspons to the `freecolor` port as well.

I would suggest this as free and used memory according to FreeBSD wiki:

Memory | Status | Reason
------------|----------|-----------
Active | Used | actively used
Inactive | Used | cannot reallocated by another process
Cached | Free | available for allocation
Free | Free | not used
Wired | Used | Not free for allocation
Buffer | --- | Don't count it as free or used ("Actual data may be stored in Active, Inactive or Wired")

This would mean that this PR should be changed that inactive memory is not counted as free memory.

What exactly do you mean with "Laundry"?

I would handle wired memory as it is done right now. It may be interesting for some users, so return it in the table, but it is used memory.

psycorama:
I've been using these wrong calculations the whole time, but never bothered to re-check them, and they don't seem to be that off …
And, yes I'm one of those people using the wired memory information. 

Regarding "Laundered Memory", ```top``` is displaying  this on my system
```Mem: 1466M Active, 1535M Inact, 1257M Laundry, 3271M Wired, 764M Buf, 185M Free```.
And ```vm.stats.vm_v_laundry_count``` does exist as well.
After some searching, if fount it was introduced with FreeBSD11.1 and does count as inactive, so we can attribute it to "Free".

mutlusun:
I looked around for laundry memory and found this information:
* http://freebsd.1045724.x6.nabble.com/PQ-LAUNDRY-td6142103.html
* https://reviews.freebsd.org/D8302
* https://lists.freebsd.org/pipermail/svn-src-user/2015-November/008494.html

Thus, I would suggest to remove the "cached" memory. As in the links above it will be removed in the future and on my computer it is always 0. Can anyone confirm this? Instead of "cached" memory we should report the "laundry" memory. Is it free or used? I haven't fully understood the two links above, but I think it should be counted as used memory.

> Introduce a new page queue, PQ_LAUNDRY, for unreferenced, i.e., inactive, dirty pages and a new thread for laundering the pages on this queue.

As far as I understand it, there is a need for "laundering" these pages, so it has to be freed before it can be used. Therefore, I would count it as used. But one could also argue that it will be freed if it is needed.

Now I would suggest this memory categorization:

Memory | Status | Reason
------------|----------|-----------
Active | Used | actively used
Inactive | Used | cannot reallocated by another process
Cached | --- | not reported by `sysctl` anymore
Free | Free | not used
Wired | Used | Not free for allocation
Buffer | --- | Don't count it as free or used ("Actual data may be stored in Active, Inactive or Wired")
Laundry | Used | Has to be laundered before it is free

psycorama:
Good Evening.

I've searched some more and revised the code once more. 
Furthermore, i pushed it into my fork at [https://github.com/psycorama/vicious](https://github.com/psycorama/vicious). I took the liberty of splitting your battery related changes into a separate commit.
You can't merge it anymore, but you could cherry-pick it into your branch if you so desire, otherwise I would suggest, merging my branch.
I do hope, the commit message explains it all.

psycorama:
The freebsd-IRC channel dug out the [FreeBSD commit for "Laundry Memory" including explanations](https://v4.freshbsd.org/commit/freebsd/src/308474) for me.

I support your proposal for counting it to used memory.

So, the only remaining free memory is real free memory.
Maybe we should inform the user about free-able memory, which would include laundry and cache. 
My system still reports cached_memory, but it is 0, too. Probably a ZFS thing, where the ARC resides in the wired memory (which is a whopping 43%, 3.5GB, of my memory, while the ARC is only about 1.2Gigs).

Mic92:
@mutlusun feel free to merge this, when it works for you.

mutlusun:
@psycorama I like your changes. Do you mind merging your local changes into my master branch? So I can merge it. Or if it is easier for you, open another pull request here? Then I will close mine.

Regarding the swap space: I would suggest to handle this in another pull request. I think I will try to parse the output of swapinfo. As I'm on travel right now I will handle this next week.

psycorama:
@mutlusun I'll clean my latest test-branch up a bit, an will create a new pull-request against Mic92:master.
Since I split your commit, you'll need to overwrite the information in your repository.

I suggest we close issues #54 and #55, then.

mutlusun:
fixed in #56

PR #56, opened 2018-03-16
FreeBSD memory widgets bugfixes and improvements

psycorama:
As discussed in #55, here is the pull request.

#### memory
* corrects calculations for amount of used and free memory 
* improves the returned table by re-arranging the four FreeBSD specific values.

  table entry | value
  --- | ---
  9 | wired memory in percent 
  10 | wired memory in megabytes
  11 | not freeable memory in percent
  12 | not freeable memory in megabytes

* removes unnecessary variables and calculations on those

#### battery
* adds "critical charging" state


mutlusun:
thanks for your contribution!

I will leave #54 open till I handled swap memory from `swapinfo`.

PR #58, opened 2018-06-06
Calculating swap space using swapinfo on FreeBSD

### LINKED ISSUE #54: FreeBSD support for memory widget w/patch

mutlusun:
Hello,
this pull request will change the memory widget to calculate swap space using `swapinfo` on FreeBSD. This should be last part to close #54.

Are there any comments to code style? Can somebody test the widget again (maybe @psycorama / @dkrand)? Thanks a lot!

### DIFF
--- a/widgets/mem_freebsd.lua
+++ b/widgets/mem_freebsd.lua
@@ -47,14 +47,34 @@ local function worker(format)
     local _swp = { buf = {}, total = nil }
 
     if vm_swap_enabled == 1 and vm_swap_total > 0 then
-        -- Get swap space in bytes
-        _swp.total = vm_swap_total
-        _swp.buf.free = _swp.total - tonumber(vm_stats.v_swapin)
+        -- Initialise variables
+        _swp.usep = 0
+        _swp.inuse = 0
+        _swp.total = 0
+        _swp.buf.free = 0
+
+        -- Read output of swapinfo
+        local f = io.popen("swapinfo")
+        -- Skip first line (table header)
+        f:read("*line")
+        -- Read content and sum up
+        for line in f:lines() do
+            local ltotal, lused, lfree = string.match(line, "%s+([%d]+)%s+([%d]+)%s+([%d]+)")
+            print(ltotal)
+            print(lused)
+            print(lfree)
+            
+            -- Add swap space in kbytes
+            _swp.total = _swp.total + tonumber(ltotal)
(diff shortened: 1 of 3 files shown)


widgets/mem_freebsd.lua
```diff
@@ -47,14 +47,34 @@ local function worker(format)
...
-        _swp.total = vm_swap_total
-        _swp.buf.free = _swp.total - tonumber(vm_stats.v_swapin)
+        -- Initialise variables
+        _swp.usep = 0
+        _swp.inuse = 0
+        _swp.total = 0
+        _swp.buf.free = 0
+
+        -- Read output of swapinfo
+        local f = io.popen("swapinfo")
```
dkrand: I'd suggest forcing swapinfo to reporting in KB, just in case the BLOCKSIZE environment variable is set. Add " -k" to the end of the command, making it:  "swapinfo -k"

widgets/mem_freebsd.lua
```diff
@@ -47,14 +47,34 @@ local function worker(format)
...
+        f:close()
+
         -- Rework into megabytes
-        _swp.total = math.floor(_swp.total/1048576)
-        _swp.buf.free = math.floor(_swp.buf.free/1048576)
+        _swp.total = math.floor(_swp.total/1024)
+        _swp.buf.free = math.floor(_swp.buf.free/1024)
         -- Calculate percentage
-        _swp.inuse = _swp.total - _swp.buf.free
         _swp.usep  = math.floor(_swp.inuse / _swp.total * 100)
```
dkrand: This is off by a factor of 1024 because _swp.inuse isn't changed to megabytes. Above you need to add:
     _swp.inuse = math.floor(_swp.inuse/1024)
to fix the percentage and to normalize inuse with the other variables.

mutlusun:
Hello,
thanks for your suggestions. I decided to read swapinfo in megabytes, so no further calculation is necessary ...

Best regards

mutlusun:
merging as there are no further comments ...

PR #59, opened 2018-09-28
Use MemAvailable for mem.free

jgkamat:
The current method of calculating the amount of free memory is
incorrect. The linux kernel provides a method of getting this value,
MemAvailable, which we can simply use directly.

See [this commit for more information](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773).

Let me know if you think this shouldn't be applied where it is currently. For me, this changes the value displayed dramatically, but it's much more correct and what I expected when installing this package.


Mic92:
Thanks

PR #61, opened 2018-10-03
net: provide list of available nets

blueyed:
This makes it easier to loop over them and/or handle them separately.


blueyed:
Should also be done for ./widgets/net_freebsd.lua then likely.

PR #63, opened 2018-10-04
update: handle stacked graphs

blueyed:
With stacked graphs each stack entry (via `stack_colors`) has its own
data point, and `data` is expected to be a table.

The `get_stack` getter was just added to awesome via [1] in [2] to
handle this case.

1: https://github.com/awesomeWM/awesome/pull/284
2: https://github.com/awesomeWM/awesome/commit/3be423fcb3c0b99d3886c57c237bde82b990cd98

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -73,7 +73,15 @@ local function update(widget, reg, disablecache)
     local function update_value(data)
         local fmtd_data = format_data(data)
         if widget.add_value ~= nil then
-            widget:add_value(tonumber(fmtd_data) and tonumber(fmtd_data)/100)
+            if widget.get_stack ~= nil and widget:get_stack() then
+                for idx, _ in ipairs(widget:get_stack_colors()) do
+                    if fmtd_data[idx] then
+                        widget:add_value(tonumber(fmtd_data[idx]) and tonumber(fmtd_data[idx]/100), idx)
+                    end
+                end
+            else
+                widget:add_value(tonumber(fmtd_data) and tonumber(fmtd_data)/100)
+            end
         elseif widget.set_value ~= nil then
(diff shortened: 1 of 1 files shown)


blueyed:
This is an old commit I had.. should be supported since awesomeWM v4.0.

init.lua
```diff
@@ -63,7 +63,15 @@ local function update(widget, reg, disablecache)
 
     local function update_value(data, t, cache)
         if widget.add_value ~= nil then
-            widget:add_value(tonumber(data) and tonumber(data)/100)
+            if widget.get_stack and widget:get_stack() then
+                for idx, col in ipairs(widget:get_stack_colors()) do
+                    if data[idx] then
+                        widget:add_value(tonumber(data[idx])/100, idx)
```
McSinyx: Shouldn't we check if `tonumber` have succeeded, i.e. `widget:add_value(tonumber(data[idx]) and tonumber(data[idx]/100), idx)`?

Please edit the patch to be compatible with current master and update the [README example on stacked graph](https://github.com/vicious-widgets/vicious/blob/065bd1d0097d6bdc737a2780fc469dd6332b0f5d/README.md#format-functions) before merging.

init.lua
```diff
@@ -63,7 +63,15 @@ local function update(widget, reg, disablecache)
 
     local function update_value(data, t, cache)
         if widget.add_value ~= nil then
-            widget:add_value(tonumber(data) and tonumber(data)/100)
+            if widget.get_stack and widget:get_stack() then
```
McSinyx: Just for style consistancy, maybe we should use `widget.get_stack ~= nil`. IMHO for safety we should additionally assert `type(widget.get_stack)` but vicious completely depends on awesome API so I'm not sure if that really matters.

Mic92:
@McSinyx can you re-test?

PR #67, opened 2018-12-14
Add more return values to mpd widget type and sec2hhmmss to helpers

### LINKED ISSUE #66: [feature request] function to convert seconds into hour,minute,second in  mpd.lua widget
could be very useful for mpd.lua widget, I added 2 more lines  as you can see:

    -- Mpd: provides Music Player Daemon information
    -- vicious.widgets.mpd
    local mpd = {}

    -- {{{ MPD widget type
    local function worker(format, warg)
    local mpd_state  = {
        ["{volume}"] = 0,
        ["{state}"]  = "N/A",
        ["{Artist}"] = "N/A",
        ["{Title}"]  = "N/A",
        ["{Album}"]  = "N/A",
        ["{Genre}"]  = "N/A",
        ["{time}"]  = "N/A",                                      <----- ADDED THIS LINE
        --["{Name}"] = "N/A",
        --["{file}"] = "N/A",
    }

    -- Fallback to MPD defaults
    local pass = warg and (warg.password or warg[1]) or "\"\""
    local host = warg and (warg.host or warg[2]) or "127.0.0.1"
    local port = warg and (warg.port or warg[3]) or "6600"

    -- Construct MPD client options
    local mpdh = "telnet://"..host..":"..port
    local echo = "echo 'password "..pass.."\nstatus\ncurrentsong\nclose'"

    -- Get data from MPD server
    local f = io.popen(echo.." | curl --connect-timeout 1 -fsm 3 "..mpdh)

    for line in f:lines() do
        for k, v in string.gmatch(line, "([%w]+):[%s](.*)$") do
            if     k == "volume" then mpd_state["{"..k.."}"] = v and tonumber(v)
            elseif k == "state"  then mpd_state["{"..k.."}"] = helpers.capitalize(v)
            elseif k == "Artist" then mpd_state["{"..k.."}"] = helpers.escape(v)
            elseif k == "Title"  then mpd_state["{"..k.."}"] = helpers.escape(v)
            elseif k == "Album"  then mpd_state["{"..k.."}"] = helpers.escape(v)
            elseif k == "Genre"  then mpd_state["{"..k.."}"] = helpers.escape(v)
            elseif k == "time"  then mpd_state["{"..k.."}"] = helpers.escape(v)   <-- AND THIS
            --elseif k == "Name" then mpd_state["{"..k.."}"] = helpers.escape(v)
            --elseif k == "file" then mpd_state["{"..k.."}"] = helpers.escape(v)
            end
        end
    end
    f:close()
 
    return mpd_state
    end
     -- }}}
 
    return setmetatable(mpd, { __call = function(_, ...) return worker(...) end })

this code allows to show time in mpd.widget in rc.lua  but in seconds, for example:

128:180    (which means track playing at 2:08 of 3:00
showing 2:08 / 3:00 or something like that would be more cool :)

McSinyx:
This should close #66.

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -136,6 +136,25 @@ function helpers.uformat(array, key, value, unit)
 end
 -- }}}
 
+-- {{{ Format playing progress
+function helpers.format_progress(elapsed, duration)
+    local em, es = elapsed / 60, elapsed % 60
+    local dm, ds = duration / 60, duration % 60
+
+    if dm < 10 then
+        return ("%d:%02d"):format(em, es), ("%d:%02d"):format(dm, ds)
+    elseif dm < 60 then
+        return ("%02d:%02d"):format(em, es), ("%02d:%02d"):format(dm, ds)
+    elseif dm < 600 then
+        return ("%d:%02d:%02d"):format(em / 60, em % 60, es),
+               ("%d:%02d:%02d"):format(dm / 60, dm % 60, ds)
+    else
+        return ("%02d:%02d:%02d"):format(em / 60, em % 60, es),
+               ("%02d:%02d:%02d"):format(dm / 60, dm % 60, ds)
+    end
+end
+-- }}}
+
 -- {{{ Escape a string
 function helpers.escape(text)
     local xml_entities = {
--- a/helpers.lua
+++ b/helpers.lua
@@ -136,25 +136,6 @@ function helpers.uformat(array, key, value, unit)
 end
 -- }}}
 
--- {{{ Format playing progress
-function helpers.format_progress(elapsed, duration)
-    local em, es = elapsed / 60, elapsed % 60
-    local dm, ds = duration / 60, duration % 60
-
-    if dm < 10 then
-        return ("%d:%02d"):format(em, es), ("%d:%02d"):format(dm, ds)
-    elseif dm < 60 then
-        return ("%02d:%02d"):format(em, es), ("%02d:%02d"):format(dm, ds)
-    elseif dm < 600 then
-        return ("%d:%02d:%02d"):format(em / 60, em % 60, es),
-               ("%d:%02d:%02d"):format(dm / 60, dm % 60, ds)
-    else
-        return ("%02d:%02d:%02d"):format(em / 60, em % 60, es),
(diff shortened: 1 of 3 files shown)


widgets/mpd_all.lua
```diff
@@ -17,39 +17,52 @@ local helpers = require("vicious.helpers")
...
     -- Get data from MPD server
-    local f = io.popen(echo.." | curl --connect-timeout 1 -fsm 3 "..mpdh)
-
+    local f = io.popen(query .. "|" .. connect)
     for line in f:lines() do
         for k, v in string.gmatch(line, "([%w]+):[%s](.*)$") do
             local key = "{" .. k .. "}"
-            if k == "volume" then
+            if k == "volume" or k == "bitrate" or
+               k == "elapsed" or k == "duration" then
```
Mic92: I think most users just want `elapsed` and `duration` to be formatted as hh:mm:ss.
Therefore I would just apply the helper for `elaspsed` and `duration` but also store
their raw values in `elapsedSeconds` and `durationSeconds`.

helpers.lua
```diff
@@ -136,6 +137,19 @@ function helpers.uformat(array, key, value, unit)
...
+function helpers.sec2hhmmss(seconds)
+    return ("%02d:%02d:%02d"):format(math.floor(seconds / 3600),
+                                     math.floor(seconds % 3600 / 60),
+                                     math.floor(seconds % 60))
+end
+
+-- {{{ Format seconds to MM:SS
+function helpers.sec2mmss(seconds)
+    return ("%02d:%02d"):format(math.floor(seconds / 60),
+                                math.floor(seconds % 60))
```
Mic92: I think skipping digits that are zero i.e. hours and minutes as sketched by https://github.com/vicious-widgets/vicious/issues/66#issuecomment-447276202 make sense to save space. (most songs will take less then an hour)

widgets/mpd_all.lua
McSinyx: To keep the table key short, I decided to use `Elapsed` and `Duration` for formatted value.

helpers.lua
McSinyx: Indeed, although elapsed and duration should have same format.

McSinyx:
Thanks for the cleaning up. I'm rebasing this now.

PR #68, opened 2019-01-15
contrib: Add hwmontemp_linux

lynix:
I'd like to introduce a new widget _hwmontemp_, which provides name-based access to hwmon
temperature sensors via sysfs.

Motivation for creating this widget was the fact that all existing hwmon widgets seem to require the sysfs path to be static. This is not the case for my system, i.e. hwmon device numbers below _/sys/class/hwmon_ are determined by module load order and vary between system boots.

This widget aims at solving this problem by making hwmon sensors accessible by name instead of sysfs path. In order to avoid unnecessary sysfs traversal for subsequent queries the name -> path mapping is cached.

### DIFF
--- a/contrib/hwmontemp_linux.lua
+++ b/contrib/hwmontemp_linux.lua
@@ -1,0 +1,61 @@
+----------------------------------------------------------------
+--   Licensed under the GNU General Public License v2
+--   (C) 2019, Alexander Koch <lynix47@gmail.com>
+----------------------------------------------------------------
+
+-- environment
+local io = { popen = io.popen, open = io.open }
+local setmetatable = setmetatable
+
+-- sysfs prefix for hwmon devices
+local sys_hwmon = "/sys/class/hwmon"
+-- cache table for hwmon device names
+local paths = {}
+
+-- transparently caching hwmon device name lookup
+function name_to_path(name)
+    if paths[name] then return paths[name] end
+
+    for sensor in io.popen("ls -1 " .. sys_hwmon):lines() do
+        local path = sys_hwmon .. "/" .. sensor
+        local f = assert(io.open(path .. "/name", "r"))
+        local sname = f:read("*line")
+        f:close()
+        if sname == name then
+            paths[name] = path
+            return path
+        end
+    end
+
+    return nil
+end
+
+-- hwmontemp: provides name-indexed temps from /sys/class/hwmon
+-- vicious.contrib.hwmontemp
+local hwmontemp_linux = {}
+
+function worker(format, warg)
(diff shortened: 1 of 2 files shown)


contrib/hwmontemp_linux.lua
```diff
@@ -0,0 +1,61 @@
...
+    if not warg[2] then
+        input = 1
+    else
+        input = warg[2]
+    end
+
+    local sensor = name_to_path(name)
+    if not sensor then return { "n/a" } end
+
+    local f = assert(io.open(sensor .. "/temp" .. input .. "_input", "r"))
```
McSinyx: IMHO string concatenation is more difficult to read than format function but that's personal preference and since you decide to put it in `contrib`, it'll be all up to you.

contrib/hwmontemp_linux.lua
```diff
@@ -0,0 +1,61 @@
...
+    name = warg[1]
+
+    if not warg[2] then
+        input = 1
+    else
+        input = warg[2]
+    end
+
+    local sensor = name_to_path(name)
+    if not sensor then return { "n/a" } end
```
McSinyx: I think it's a convention in this repo to use `N/A`.

contrib/hwmontemp_linux.lua
```diff
@@ -0,0 +1,61 @@
...
+
+    return nil
+end
+
+-- hwmontemp: provides name-indexed temps from /sys/class/hwmon
+-- vicious.contrib.hwmontemp
+local hwmontemp_linux = {}
+
+function worker(format, warg)
+    if not warg[1] then
```
McSinyx: Do you think it'd be better to assert `warg`'s type to pass error in silence?

contrib/hwmontemp_linux.lua
lynix: Well, I thought string concatenation to be more efficient than parsing format strings... but maybe readability is more important than saving a few cycles ;)

I thought `contrib` to be something like the staging folder for new plugins. If it's possible to go with `widgets` directly I'd happy to put it there directly (so I don't need to import `contrib` in my awesome config).

Guess I'll better open a new pull request for this as there need to be some changes to my patch, right?

contrib/hwmontemp_linux.lua
lynix: Thanks, will fix this.

contrib/hwmontemp_linux.lua
lynix: Yes you're right: this way the user even gets a popup when using the widget in awesome. Will fix this.

Issue #10, opened 2013-10-30
[Supported by Awesome with playerctl] Widget type for MPRIS-compliant players


kierun:
Would it be possible to create a clone of the mpd widget but supporting xmms2?

I would offer to do it but my lua skills are weak.  I do volunteer to test it though.

PR #70, opened 2019-01-15
widgets: Add hwmontemp_linux

lynix:
Improved version of #68 with review comments taken into account.
- Switch from `contrib` to `widgets`
- Use `s:format()` instead of string concatenation for input file
- Adhere to `"N/A"` convention
- Use assertion for worker argument

Original pull description:

> I'd like to introduce a new widget hwmontemp, which provides name-based access to hwmon temperature sensors via sysfs.
> 
> Motivation for creating this widget was the fact that all existing hwmon widgets seem to require the sysfs path to be static. This is not the case for my system, i.e. hwmon device numbers below /sys/class/hwmon are determined by module load order and vary between system boots.
> 
> This widget aims at solving this problem by making hwmon sensors accessible by name instead of sysfs path. In order to avoid unnecessary sysfs traversal for subsequent queries the name -> path mapping is cached.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,22 @@ Supported platforms: GNU/Linux, requiring `hddtemp` and `curl`.
 * Returns a table with string keys, using hard drives as a base, e.g.
   `${/dev/sda}` and `${/dev/sdc}`.
 
+### vicious.widgets.hwmontemp
+
+Provides name-based access to hwmon devices via sysfs.
+
+Supported platforms: GNU/Linux
+
+* Argument: table with sensor name and (optional) input number, e.g.
+  `{ "radeon", 2 }` (input no. is assumed to be 1 if omitted)
+* Returns a table with just the temperature value: `$1`
+* Usage example:
+
+```lua
+gputemp = wibox.widget.textbox()
+vicious.register(gputemp, vicious.contrib.hwmontemp, " $1°C", 5, { "radeon" })
+```
+
 ### vicious.widgets.mbox
 
 Provides the subject of last e-mail in a mbox file.
--- a/widgets/hwmontemp_linux.lua
+++ b/widgets/hwmontemp_linux.lua
@@ -1,0 +1,60 @@
+----------------------------------------------------------------
+--   Licensed under the GNU General Public License v2
+--   (C) 2019, Alexander Koch <lynix47@gmail.com>
+----------------------------------------------------------------
+
+-- environment
+local io = { popen = io.popen, open = io.open }
+local assert = assert
+local setmetatable = setmetatable
+
+-- sysfs prefix for hwmon devices
+local sys_hwmon = "/sys/class/hwmon/"
+-- cache table for hwmon device names
+local paths = {}
+
+-- transparently caching hwmon device name lookup
+function name_to_path(name)
+    if paths[name] then return paths[name] end
+
+    for sensor in io.popen("ls -1 " .. sys_hwmon):lines() do
+        local path = sys_hwmon .. sensor
+        local f = assert(io.open(path .. "/name", "r"))
+        local sname = f:read("*line")
+        f:close()
+        if sname == name then
+            paths[name] = path
+            return path
+        end
+    end
+
+    return nil
+end
+
+-- hwmontemp: provides name-indexed temps from /sys/class/hwmon
+-- vicious.widgets.hwmontemp
+local hwmontemp_linux = {}
+
+function worker(format, warg)
+    assert(type(warg) == "table", "invalid hwmontemp argument: must be a table")
+    name = warg[1]
+
+    if not warg[2] then
+        input = 1
+    else
+        input = warg[2]
+    end
+
+    local sensor = name_to_path(name)
+    if not sensor then return { "N/A" } end
+
+    local f = assert(io.open(("%s/temp%d_input"):format(sensor, input), "r"))
+    local temp = f:read("*line")
+    f:close()
+
+    return { temp / 1000 }
+end
+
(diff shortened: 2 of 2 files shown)


widgets/hwmontemp_linux.lua
```diff
@@ -0,0 +1,60 @@
...
+-- sysfs prefix for hwmon devices
+local sys_hwmon = "/sys/class/hwmon/"
+-- cache table for hwmon device names
+local paths = {}
+
+-- transparently caching hwmon device name lookup
+function name_to_path(name)
+    if paths[name] then return paths[name] end
+
+    for sensor in io.popen("ls -1 " .. sys_hwmon):lines() do
```
Mic92: There is also `pathstotable` which does this.

widgets/hwmontemp_linux.lua
```diff
@@ -0,0 +1,60 @@
...
+    if not warg[2] then
+        input = 1
+    else
+        input = warg[2]
+    end
+
+    local sensor = name_to_path(name)
+    if not sensor then return { "N/A" } end
+
+    local f = assert(io.open(("%s/temp%d_input"):format(sensor, input), "r"))
```
Mic92: How is this widget different from `thermal_linux.lua`?

widgets/hwmontemp_linux.lua
```diff
@@ -0,0 +1,60 @@
...
+    assert(type(warg) == "table", "invalid hwmontemp argument: must be a table")
+    name = warg[1]
+
+    if not warg[2] then
+        input = 1
+    else
+        input = warg[2]
+    end
+
+    local sensor = name_to_path(name)
```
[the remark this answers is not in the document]
lynix: Well, it _could_. But `thermal_linux` does not offer input selection for the `hwmon` zone, and name caching would not fit in there so nicely either as this only makes sense for `hwmon` and not `core` or `proc`.

It's up to you. If you don't like the approach of having this new widget then please feel free to close this pull request. I don't know if I'd be coming up with another one that extends `thermal_linux` in this case, depends on the amount spare-time I have.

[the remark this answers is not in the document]
McSinyx: Would `vicious.widgets.thermal{'radeon', 'hwmon', 'temp1_input'}` give the desired result?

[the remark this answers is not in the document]
Mic92: well it would be still possible to keep backwards compatibility by still accepting the old format, but removing it from documentation and then just check in lua if the string starts with `/`.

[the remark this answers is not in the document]
lynix: > (...) it'd be a lot better if we can have similar behaviour to the FreeBSD variant (input of absolute path to the file only (...)

That's _exactly_ what I do not want, as paths do change randomly in case of hwmon sensors. They depend on the device number, which in turn depends on kernel module load order. This is not deterministic, and the only reason why I have written my new widget (as stated on top).

widgets/hwmontemp_linux.lua
```diff
@@ -0,0 +1,60 @@
...
+-- transparently caching hwmon device name lookup
+function name_to_path(name)
+    if paths[name] then return paths[name] end
+
+    for sensor in io.popen("ls -1 " .. sys_hwmon):lines() do
+        local path = sys_hwmon .. sensor
+        local f = assert(io.open(path .. "/name", "r"))
+        local sname = f:read("*line")
+        f:close()
+        if sname == name then
```
McSinyx: I'm so sorry that I missed this part (I guess the reason is I was staying up late studying for finals). Also as @mic92 said, there is [`helpers.pathtotable`](https://github.com/vicious-widgets/vicious/blob/6b4910e0d14f97edebd2ab43698e63b1152bfad9/helpers.lua#L96) that you can use instead of manually `ls`ing.

lynix:
```lua suggestion
        local sensors = helpers.pathtotable("/sys/class/hwmon")
        local i = 0
        while sensors["hwmon" .. i] do
            local s = "hwmon" .. i
            if sensors[s].name == name .. "\n" then
```
Doesn't seem much easier to me, but maybe I'm just not getting the intended way to use `helpers.pathtotable` ;)

McSinyx:
@Mic92: I think I am merging this. IMHO the most important purpose of the project is to provide a more flexible method of handling updates of files/command output (widget registration, handy formatter, caching and forcing updates). `widgets/` are just examples of how things could be done, and there isn't any perfect solution to fetch information, especially when things are changing rapidly. Since I'm quite certain that I can maintain this widget type, I think it is sensible to include it in the repo.

Not exactly related to this PR, but I'm thinking about adding a few widget types, namely for pulseaudio and MPRIS-supported players. The later one could solve #10.

Mic92:
@McSinyx I don't have a strong opinion on this. I just want to make sure we don't add too much duplication by adding widgets that make more or less the same.

README.md
```diff
@@ -259,6 +259,22 @@ Supported platforms: GNU/Linux, requiring `hddtemp` and `curl`.
...
+Supported platforms: GNU/Linux
+
+* Argument: table with sensor name and (optional) input number, e.g.
+  `{ "radeon", 2 }` (input no. is assumed to be 1 if omitted)
+* Returns a table with just the temperature value: `$1`
+* Usage example:
+
+```lua
+gputemp = wibox.widget.textbox()
+vicious.register(gputemp, vicious.contrib.hwmontemp, " $1°C", 5, { "radeon" })
```
McSinyx: I'm so sorry for the delay--lunar new year is an important holiday for families to gather in my country. Please update this example to `vicious.widgets.hwmontemp` and I will merge. Thank you and @Mic92, and happy new year!

PR #72, opened 2019-01-21
Fix luacheck issues and add .luacheckrc

blueyed:
Uses `reg.timeout` for clarity (triggered via "shadowing upvalue timer
on line 13"), but keeps `reg.timer` for backward compatibility.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,0 +1,9 @@
+-- Only allow symbols available in all Lua versions
+std = "min"
+
+-- Global objects defined by the C code
+read_globals = {
+    "timer",  -- deprecated, but used in older versions.
+}
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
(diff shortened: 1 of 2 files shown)


.luacheckrc
```diff
@@ -0,0 +1,9 @@
+-- Only allow symbols available in all Lua versions
+std = "min"
+
+-- Global objects defined by the C code
+read_globals = {
+    "timer",  -- deprecated, but used in older versions.
+}
+
+-- vim: filetype=lua:expandtab:shiftwidth=4:tabstop=8:softtabstop=4:textwidth=80
```
blueyed: This is based on `.luacheckrc` from awesome itself.

McSinyx: Is luacheck supposed to be used manually only or are we going to add to add a CI like Travis? And can you merge the PR afterwards, since this is already approved by @Mic92. I feel uncomfortable that while I'm neither the only one with write access nor the one with most knowledge or experience, I keep merging other members' patches.

PR #75, opened 2019-03-02
[mpd] Use math.floor when formatting song progress

JuanKman94:
There was an error in the `format_progress` function in my environment.

```bash
me@localhost vicious> lua -v
Lua 5.3.5  Copyright (C) 1994-2018 Lua.org, PUC-Rio
me@localhost vicious> awesome -v
awesome v4.3-91-ga0731023 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

PR #76, opened 2019-03-06
Add battery support for OpenBSD

kiike:
This PR adds battery support for OpenBSD. I have been using for several months already so I decided it would be a good idea to request its addition here. I plan on adding some more widgets for OpenBSD as needed.

### DIFF
--- a/widgets/bat_openbsd.lua
+++ b/widgets/bat_openbsd.lua
@@ -27,7 +27,6 @@ local function worker(format, warg)
 	    bat_info[key] = value
         end
     end
-    proc:close()
 
     -- current state
     local states = {[0] = "↯", -- high
--- a/widgets/bat_openbsd.lua
+++ b/widgets/bat_openbsd.lua
@@ -1,0 +1,74 @@
+-- {{{ Grab environment
+local setmetatable = setmetatable
+local tonumber = tonumber
+local string = { format = string.format, gmatch = string.gmatch }
+local io = { popen = io.popen }
+local math = { floor = math.floor }
+local helpers = require("vicious.helpers")
+local string = {
+    gmatch = string.gmatch,
+    match = string.match,
+    format = string.format
+
+}
+-- }}}
+local bat_openbsd = {}
+
+local function worker(format, warg)
+    local battery = warg or "bat0"
+    local filter = string.format("hw.sensors.acpi%s", battery)
+    local cmd = string.format("sysctl -a | grep '^%s'", filter)
+    local proc = io.popen(cmd)
+
+    local bat_info = {}
+    for line in proc:lines("*l") do
+        for key, value in string.gmatch(line, "(%S+)=(%S+)") do
+	    key = key:gsub(filter .. ".", "")
+	    bat_info[key] = value
+        end
+    end
+    proc:close()
+
+    -- current state
+    local states = {[0] = "↯", -- high
+                    [1] = "-", -- low
+                    [2] = "!", -- critical
+                    [3] = "+", -- charging
+                    [4] = "n/a", -- critical
+                    [255] = "⌁"} -- unknown
+    local state = states[tonumber(bat_info.raw0)]
+
+    -- battery capacity in percent
+    local percent = tonumber(bat_info.watthour3 / bat_info.watthour0 * 100)
+
+    local time
+    if tonumber(bat_info.power0) < 1 then
+        time = "∞"
+    else
+	local raw_time = bat_info.watthour3 / bat_info.power0
+	local hours = math.floor(raw_time)
+	local minutes = raw_time % 1
+	time = string.format("%d:%0.2d", hours, minutes)
+    end
+
+    -- calculate wear level from (last full / design) capacity
+    local wear = "N/A"
+    if bat_info.watthour0 and bat_info.watthour4 then
+        local l_full = tonumber(bat_info.watthour0)
+        local design = tonumber(bat_info.watthour4)
+        wear = math.floor(l_full / design * 100)
+    end
+
+    -- dis-/charging rate as presented by battery
+    local rate = bat_info.power0
+
+    -- returns
+    --  * state (high "↯", discharging "-", charging "+", N/A "⌁" }
+    --  * remaining_capacity (percent)
+    --  * remaining_time, by battery
+    --  * wear level (percent)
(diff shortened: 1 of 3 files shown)


widgets/bat_openbsd.lua
```diff
@@ -0,0 +1,74 @@
...
+    end
+    proc:close()
+
+    -- current state
+    local states = {[0] = "↯", -- high
+                    [1] = "-", -- low
+                    [2] = "!", -- critical
+                    [3] = "+", -- charging
+                    [4] = "n/a", -- critical
+                    [255] = "⌁"} -- unknown
```
McSinyx: I think `N/A` (conventionally this repo uses capitalized version) is more suitable for *unknown* state.

widgets/bat_openbsd.lua
```diff
@@ -0,0 +1,74 @@
...
+    -- dis-/charging rate as presented by battery
+    local rate = bat_info.power0
+
+    -- returns
+    --  * state (high "↯", discharging "-", charging "+", N/A "⌁" }
+    --  * remaining_capacity (percent)
+    --  * remaining_time, by battery
+    --  * wear level (percent)
+    --  * present_rate (W)
+    return {state, percent, time, wear, rate}
```
McSinyx: This should be documented in `README.md`.

widgets/bat_openbsd.lua
```diff
@@ -0,0 +1,74 @@
...
+local string = { format = string.format, gmatch = string.gmatch }
+local io = { popen = io.popen }
+local math = { floor = math.floor }
+local helpers = require("vicious.helpers")
+local string = {
+    gmatch = string.gmatch,
+    match = string.match,
+    format = string.format
+
+}
```
McSinyx: IMHO it'd look better if you switch line 12 and 13.

widgets/bat_openbsd.lua
```diff
@@ -0,0 +1,74 @@
...
+    local proc = io.popen(cmd)
+
+    local bat_info = {}
+    for line in proc:lines("*l") do
+        for key, value in string.gmatch(line, "(%S+)=(%S+)") do
+	    key = key:gsub(filter .. ".", "")
+	    bat_info[key] = value
+        end
+    end
+    proc:close()
```
McSinyx: `io.lines` closes the file after the iteration completes.

widgets/bat_openbsd.lua
```diff
@@ -0,0 +1,74 @@
...
+    format = string.format
+
+}
+-- }}}
+local bat_openbsd = {}
+
+local function worker(format, warg)
+    local battery = warg or "bat0"
+    local filter = string.format("hw.sensors.acpi%s", battery)
+    local cmd = string.format("sysctl -a | grep '^%s'", filter)
```
McSinyx: I acknowledge that this is totally up to personal preference but maybe we can join these two lines?

widgets/bat_openbsd.lua
kiike: Done in 7d62dd3. I also reworded the comments.

widgets/bat_openbsd.lua
kiike: I just added it to the README. The returned table is the same as in the Linux and FreeBSD counterparts (I actually use the same config for both my OpenBSD and my Linux rigs).

widgets/bat_openbsd.lua
kiike: Definitely! I don't know how I couldn't see that line.

widgets/bat_openbsd.lua
kiike: TIL! Well, one less line to maintain.

widgets/bat_openbsd.lua
kiike: Since I reuse the filter variable later in the function, I'd rather keep it. I can get rid of the `cmd` one if you want but, IMHO, that function, as it is, is pretty easy to follow.

kiike:
Thanks a lot for the comments. I have added some commits to see if it better qualifies for inclusion.

widgets/bat_openbsd.lua
McSinyx: > I reuse the filter variable later in the function.

I'm so sorry that I missed that. Yes, please keep it that way.

widgets/bat_openbsd.lua
McSinyx: > I actually use the same config for both my OpenBSD and my Linux rigs

Somehow this makes me extra happy today!

McSinyx:
The PR looks all good to me now, I'm merging it right the way! We're looking forward to seeing other OpenBSD widget type of yours (if you're dedicated, just ask for an invitation to this so-called *organization*: AFAIK none of the current members uses OpenBSD). Thank you for your contribution.

kiike:
Thank you for the constructive attitude and the welcoming feedback! I will definitely contribute more code as soon as I can!

widgets/bat_openbsd.lua
```diff
@@ -0,0 +1,72 @@
...
+    local percent = tonumber(bat_info.watthour3 / bat_info.watthour0 * 100)
+
+    local time
+    if tonumber(bat_info.power0) < 1 then
+        time = "∞"
+    else
+	local raw_time = bat_info.watthour3 / bat_info.power0
+	local hours = math.floor(raw_time)
+	local minutes = raw_time % 1
+	time = string.format("%d:%0.2d", hours, minutes)
```
McSinyx: Relating to https://github.com/vicious-widgets/vicious/pull/75#issuecomment-471214820, could you please also floor minutes for Lua 5.3 compatibility? Or you could explain why is `minutes = raw_time % 1` so that I have to confidence to do the change myself.

kiike: Thanks again for your feedback. I've now addressed your comments in #78. Line 48 shoud've read `... 60 * (raw_time % 1). That is fixed in #78, along with some cruft that was detected with `luacheck`.

widgets/bat_openbsd.lua
McSinyx: My mistake: while `io.lines(file)` does closes `file` after the iteration, `file:lines()` does not. I will fix this along with the refactoring to completely get rid of `popen`.

Issue #74, opened 2019-02-26
Stop using io.popen


psychon:
This is a forwarding of https://github.com/awesomeWM/awesome/issues/2697.

The short version: The volume widget uses `io.popen` to start `amixer`. For some unknown reason, `amixer` hangs(*). (Can be seen in the output of `pstree`: There is still an `amixer` process there.) As a result, all of AwesomeWM hangs.

Please use `awful.spawn.easy_async` instead.

(*) Well, or alternatively: For some weird reason, `amixer` exits, becomes a zombie (hence still shows up in `pstree`), but is not reaped via `pclose`. I'd need a backtrace for the awesome C code to see if is hung in reading `amixer`'s output or in the `pclose`, but I would expect that reading the output hangs.

CC @furgerf

Mic92:
makes sense.

psychon:
One more adventage of `awful.spawn` is that it also supports "arrays" instead of strings and no shell is involved. Specifically, this code:
```lua
    local cmd = "amixer -M get "
    for _, arg in ipairs(warg) do
      cmd = cmd .. " " .. helpers.shellquote(arg)
    end

    -- Get mixer control contents
    local f = io.popen(cmd)
```
can be replaced with
```lua
    local cmd = { "amixer", "-M", "get" }
    for _, arg in ipairs(warg) do
        table.insert(cmd, arg)
    end
   awful.spawn.easy_async(cmd, some_callback_function)
```
AwesomeWM's C code does not have to do word splitting to turn a string into an array, but directly uses the provided array. Thus, spaces and quotation marks are no longer a problem (well and even better: things like `$(rm -rf /)` do not cause problems (unless you happen to sent this through a shell, e.g. `{"sh","-c","echo $(rm -rf /)"}`)).

McSinyx:
IMHO the solution might need to be more complex than simply replacing `io.popen` with Awesome `spawn.*`, since Vicious can currently be used as a standalone library we better not break backward compatibility. I'm thinking about creating a custom wrapper in `helpers` that fallbacks to `io.popen` in case Awesome API is not exposed.

Slightly related to this, all IO-intensive (heavy grepping or fetching from the Internet) widget types should make use of Awesome asynchronous when possible.

mutlusun:
Do we have any information about our users? Is there anyone who uses vicious outside of awesome? I wouldn't be sure that there is someone. Maybe we can assume that the awesome API is available.

[...]

PR #77, opened 2019-03-08
update bat_freebsd

mutlusun:
Hello,

thanks to the discussion about the openbsd battery widget I updated the freebsd widget. I'm happy about your comments!

Sorry that I do not invest that much time on vicious right now ...

Best
mutlusun

### DIFF
--- a/widgets/bat_freebsd.lua
+++ b/widgets/bat_freebsd.lua
@@ -8,8 +8,8 @@ local string = {
     gmatch = string.gmatch,
     match = string.match,
     format = string.format
-
 }
+
 -- }}}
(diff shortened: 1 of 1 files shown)


McSinyx:
This looks great to me: I think you can merge it now.

> Sorry that I do not invest that much time on vicious right now...

You shouldn't be sorry. As a student I have a lot more free time than you guys (that's why I offered a helping hand), and Vicious is pretty stable after a decade of development so there isn't much work. The only pending task at the moment is #74 which require refactoring most widget types to use Awesome `awful.spawn` when possible, but I haven't think of a way to do that yet.

PR #78, opened 2019-03-10
Fix bat_openbsd compatibility with Lua 5.3 and remaining time display

kiike:
This PR fixes compatibility with Lua 5.3. I didn't realise it would crash on this version because the OpenBSD Awesome port runs in 5.2. I should've checked beforehand, so let me apologise for not keeping compatibility. I've also changed the way I get hours and minutes from the sysctl output: I was using the `% 1` operation to get the fractional part of a number but I now grab both hours and minutes from `math.modf` instead.


kiike:
I just localized `math.modf`. Thanks for your review.

McSinyx:
Thanks, now I'm able to publish a new release to correct a mistake I made in the previous one :stuck_out_tongue_winking_eye:

PR #79, opened 2019-03-15
[mpd] More math.floor for format progress fix

lgaggini:
Hi,

I add some more math.floor to fix bad argument to 'format' (number has no integer representation) error.


McSinyx:
Thanks, we had definitely missed that.

PR #80, opened 2019-03-15
[mpd] Playing progress percentage

lgaggini:
Hi,

inspired from the progress percentage indicator in mpd mpc client I add the Progress percentage field to mpd_state.

### DIFF
--- a/widgets/mpd_all.lua
+++ b/widgets/mpd_all.lua
@@ -43,6 +43,14 @@ function format_progress(elapsed, duration)
 end
 -- }}}
 
+-- {{{ Format playing progress (percentage)
+function format_progress_percentage(elapsed, duration)
+    local percentage = math.floor((elapsed / duration) * 100)
+    
+    return ("%d%%"):format(percentage)
+end
+-- }}}
+
 -- {{{ MPD widget type
 local function worker(format, warg)
     -- Fallback values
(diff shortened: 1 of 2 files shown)


widgets/mpd_all.lua
```diff
@@ -43,6 +43,14 @@ function format_progress(elapsed, duration)
 end
 -- }}}
 
+-- {{{ Format playing progress (percentage)
+function format_progress_percentage(elapsed, duration)
+    local percentage = math.floor((elapsed / duration) * 100)
```
McSinyx: Do you think it would be more precise if `math.floor((elapsed / duration) * 100 + 0.5)` was used?

lgaggini:
I add also a commit to prevent division by zero in the progress percentage calculation.

widgets/mpd_all.lua
lgaggini: Fair point, not sure about what approach is more precise.

lgaggini: I prefer your version, for example in case of a percentage of 25.9% I like to have 26% instead of 25%.

```
> 200/770
0.2597
> math.floor((200/770)*100)
25
> math.floor((200/770)*100+0.5)
26
```

McSinyx:
It looks perfect to me now. Let's merge!

PR #81, opened 2019-03-18
[mpd] fixes if-else block missing end

lgaggini:
Too bad in the last commit I missed an end keyword, sorry.


McSinyx:
No worries, I missed that when reviewing the #80 too :smile:

PR #83, opened 2019-05-20
Surface {when} in UTC for weather widget

frioux:
Here's how I use this:

```lua
   -- copied from http://lua-users.org/wiki/TimeZone
   local function get_timezone_offset()
       local ts = os.time()
       local utcdate   = os.date("!*t", ts)
       local localdate = os.date("*t", ts)
       localdate.isdst = false -- this is the trick
       return os.difftime(os.time(localdate), os.time(utcdate))
   end

   -- ...

   vicious.register(actual_widget, vicious.widgets.weather,
     function (widget, args)
       actual_tooltip:set_text(
         "City: " .. args["{city}"] ..
           "\nWind: " .. args["{windmph}"] .. "mph " ..
           "\nSky: " .. args["{sky}"] ..
           "\nHumidity: " .. args["{humid}"] ..
           "\nMeasured at: " .. os.date("%F %T", args["{when}"] + get_timezone_offset()))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -505,7 +505,7 @@ Supported platforms: any having Awesome and `curl` installed.
 * Argument: the ICAO station code, e.g. `"LDRI"`
 * Returns a table with string keys: `${city}`, `${wind}`, `${windmph}`,
   `${windkmh}`, `${sky}`, `${weather}`, `${tempf}`, `${tempc}`, `${humid}`,
-  `${dewf}`, `${dewc}` and `${press}`
+  `${dewf}`, `${dewc}` and `${press}`, `${when}`
 
 ### vicious.widgets.wifi
 
--- a/widgets/weather_all.lua
+++ b/widgets/weather_all.lua
@@ -6,6 +6,7 @@
 -- {{{ Grab environment
 local tonumber = tonumber
 local math = { ceil = math.ceil }
+local os = { date = os.date, difftime = os.difftime, time = os.time }
 local string = { match = string.match }
 
 local spawn = require"vicious.spawn"
@@ -17,12 +18,22 @@ local helpers = require"vicious.helpers"
 -- vicious.widgets.weather
 local weather_all = {}
 
+-- copied from http://lua-users.org/wiki/TimeZone
+local function get_timezone_offset()
+    local ts = os.time()
+    local utcdate   = os.date("!*t", ts)
+    local localdate = os.date("*t", ts)
+    localdate.isdst = false -- this is the trick
+    return os.difftime(os.time(localdate), os.time(utcdate))
+end
+
 
 -- {{{ Weather widget type
 local function parse(stdout, stderr, exitreason, exitcode)
     -- Initialize function tables
     local _weather = {
         ["{city}"]    = "N/A",
+        ["{when}"]    = "N/A",
         ["{wind}"]    = "N/A",
         ["{windmph}"] = "N/A",
         ["{windkmh}"] = "N/A",
@@ -57,6 +68,19 @@ local function parse(stdout, stderr, exitreason, exitcode)
        string.match(stdout, "Relative[%s]Humidity:[%s]([%d]+)%%") or _weather["{humid}"]
     _weather["{press}"]   = -- Pressure in hPa
        string.match(stdout, "Pressure[%s].+%((.+)[%s]hPa%)") or _weather["{press}"]
+
+    local year, month, day, hour, min =
+        string.match(stdout, "(%d%d%d%d).(%d%d).(%d%d) (%d%d)(%d%d) UTC")
+    if year ~= nil then
+       local utctable = {
+           year = year,
+           month = month,
+           day = day,
+           hour = hour,
+           min = min,
+       }
+       _weather["{when}"] = os.time(utctable) + get_timezone_offset()
+    end
 
     -- Wind speed in km/h if MPH was available
     if _weather["{windmph}"] ~= "N/A" then


widgets/weather_all.lua
```diff
@@ -58,6 +58,19 @@ local function parse(stdout, stderr, exitreason, exitcode)
...
+
+    local utctable = {
+        year = year,
+        month = month,
+        day = day,
+        hour = hour,
+        min = min,
+    }
+
+    _weather["{when}"] = os.time(utctable)
```
McSinyx: Please localize `os.time`, which might allow the code to run a bit faster. It might not matter much on current computers, but that's what I see in most Lua projects--we might be simply paranoid though.

Additionally, have you considered the case when `stdout` is abnormal, which could be due to server maintenance or wrong station code? If that happens, fallback to `N/A`.

On coding style, I'd prefer to have trailing newlines 63 and 71 removed, since this block of code forms a logical section. It's, however, mostly because the function is already tooo long (mea culpa) and thus blank line would be quite distracting.

McSinyx: Please localize `os.time`, which might allow the code to run a bit faster. It might not matter much on current computers, but that's what I see in most Lua projects--we might be simply paranoid though.

Additionally, have you considered the case when stdout is abnormal, which could be due to server maintenance or wrong station code? If that happens, fallback to `N/A`.

On coding style, I'd prefer to have trailing newlines 63 and 71 removed, since this block of code forms a logical section. It's, however, mostly because the function is already tooo long (mea culpa) and thus blank line would be quite distracting.

README.md
```diff
@@ -505,7 +505,7 @@ Supported platforms: any having Awesome and `curl` installed.
 * Argument: the ICAO station code, e.g. `"LDRI"`
 * Returns a table with string keys: `${city}`, `${wind}`, `${windmph}`,
   `${windkmh}`, `${sky}`, `${weather}`, `${tempf}`, `${tempc}`, `${humid}`,
-  `${dewf}`, `${dewc}` and `${press}`
+  `${dewf}`, `${dewc}` and `${press}`, `${when}`
```
McSinyx: Whilst we appreciate to get extra information, IMHO it's quite cumbersome to use it. Do you think it'd be more convenient to receive the relative offset, or at least local time? As this would involve adding `get_timezone_offset` to `weather_all`, I'd like to add that `os.date` by default use current time (i.e. the one returned by `os.time`).

widgets/weather_all.lua
frioux: I'll take care of all of these; thanks for explaining some of them, I'd always wondered about the localization!

README.md
frioux: Sure!  Honestly I was a little concerned that doing all the offset lookup etc when the user isn't using this field would be a bunch of wasted work.  If you don't mind always translating it to local time that's fine with me.

I still think surfacing it as this data type is best since then the user can surface it as a duration (`when - os.time()`) or some other format (`os.date`) if they want to without too much hassle

frioux:
Regarding

> BTW, I'm curious about the way you use the widget type: why don't you register actual_tooltip directly, since set_text will eventually be called by Vicious.

Any stylistic errors are probably me just not knowing better.  Here's the full text of the widget creation function, if you want to tell me how I could do it in a less silly way:

```lua
local weather_widget = function(
  code,
  url,
  name
)
  local actual_widget = wibox.widget.textbox()
  local actual_tooltip = awful.tooltip({ objects = { actual_widget } });

  actual_widget:buttons(awful.util.table.join(
    awful.button({}, 1, function () awful.spawn("firefox '" .. url .. os.time() .. "'") end)
  ))

  vicious.register(actual_widget, vicious.widgets.weather,
    function (widget, args)
      actual_tooltip:set_text(
        "City: " .. args["{city}"] ..
          "\nWind: " .. args["{windmph}"] .. "mph " ..
          "\nSky: " .. args["{sky}"] ..
          "\nHumidity: " .. args["{humid}"] ..
          "\nMeasured at: " .. os.date("%F %T", args["{when}"] + get_timezone_offset()))
      return name .. " " .. args["{tempf}"] .. "°F"
  end, 60 * 10, code)

  return actual_widget
end
```

widgets/weather_all.lua
frioux: The `N/A` fallback is a little tricky.  If we surface the information as seconds since the epoch, the user will have to check for `N/A` to guard against passing non-integers to `os.date` (or whatever.)  On the other hand, if we pass it as a zero, then the user has to check for a zero to render it as `N/A` themselves.  Ideas?

README.md
frioux: (Done)

widgets/weather_all.lua
McSinyx: As you can see at the beginning of `parse`, `_weather` was initiated with `N/A`. I'm just trying to follow this convention. IMHO if users gotta check, then `nil` would be most Lua-ish, so I guess it was because both numbers and `N/A` can be lazily and prettily formatted with `%s`.

widgets/weather_all.lua
```diff
@@ -17,6 +18,15 @@ local helpers = require"vicious.helpers"
 -- vicious.widgets.weather
 local weather_all = {}
 
+-- copied from http://lua-users.org/wiki/TimeZone
+local function get_timezone_offset()
+    local ts = os.time()
+    local utcdate   = os.date("!*t", ts)
+    local localdate = os.date("*t", ts)
+    localdate.isdst = false -- this is the trick
```
McSinyx: I've just (mentally) evaluate this and realize that this is not supposed to be used here. The original function was to compute timezone, whilst what we want is actual offset. To sum up all requests, I drafted a patch as follow:

```diff
diff --git a/widgets/weather_all.lua b/widgets/weather_all.lua
index 6bd4cfc..ba7e886 100644
--- a/widgets/weather_all.lua
+++ b/widgets/weather_all.lua
@@ -6,6 +6,7 @@
 -- {{{ Grab environment
 local tonumber = tonumber
 local math = { ceil = math.ceil }
+local os = { date = os.date, difftime = os.difftime, time = os.time }
 local string = { match = string.match }
 
 local spawn = require"vicious.spawn"
@@ -17,12 +18,15 @@ local helpers = require"vicious.helpers"
 -- vicious.widgets.weather
 local weather_all = {}
 
+-- Actual time offset with day-saving taken onto account
+local TIME_OFFSET = os.difftime(os.time, os.time(os.date"!*t"))
 
 -- {{{ Weather widget type
 local function parse(stdout, stderr, exitreason, exitcode)
     -- Initialize function tables
     local _weather = {
         ["{city}"]    = "N/A",
+        ["{when}"]    = "N/A",
         ["{wind}"]    = "N/A",
         ["{windmph}"] = "N/A",
         ["{windkmh}"] = "N/A",
@@ -58,6 +62,18 @@ local function parse(stdout, stderr, exitreason, exitcode)
     _weather["{press}"]   = -- Pressure in hPa
        string.match(stdout, "Pressure[%s].+%((.+)[%s]hPa%)") or _weather["{press}"]
 
+    local year, month, day, hour, min =
+        string.match(stdout, "(%d%d%d%d).(%d%d).(%d%d) (%d%d)(%d%d) UTC")
+    if year ~= nil then
+        _weather["{when}"] = TIME_OFFSET + os.time{
+            year = year,
+            month = month,
+            day = day,
+            hour = hour,
+            minute = minute
+        }
+    end
+
     -- Wind speed in km/h if MPH was available
     if _weather["{windmph}"] ~= "N/A" then
        _weather["{windmph}"] = tonumber(_weather["{windmph}"])
```

Day saving is not available in my tropical country, so feel free to discuss if you think I'm wrong.

frioux: Right; this will mean the offset will be wrong until users restart awesome at DST.  If you are ok with that it's fine with me.

McSinyx: I totally missed that d-: My point was that `isdst` should not be changed, and my attempt to save a few instructions failed miserably. I guess `time_offset` still needs to be a function, i.e.

```lua
local function time_offset()
    return os.difftime(os.time(), os.time(os.date"!*t"))
end
```

frioux: yep that seems good to me.  You can apply this change or I can modify my patch; whichever you prefer.

McSinyx: > I can modify my patch

Please do, then we can squash your commits together and it'd make a more meaningful `blame`.

McSinyx:
Your widget creation is totally fine; I thought the formatting function didn't return anything and thus asked about it.

frioux:
OK, I squashed and incorporated most of your suggestions.  The one thing I didn't take was your `time_offset` function; it was actually an hour off, and based on the comment from the code I copy pasted I assume it is related to DST.

McSinyx:
If it works for you it means that it correct. I'm merging this now.

frioux:
Woo, thanks!

McSinyx:
You're welcome, although I'm the one should be thankful of your work. If you have any improvement or suggestion in the future, don't hesitate to contribute to this project.

PR #85, opened 2019-07-18
BSSID string in wifiiw_linux module

xaverh:
Access the BSSID via `${bssid}` in `Connected to ff:ff:ff:00:00:00 (on wlan0)` string in `iw dev wlan0 link` output. (wifiiw_linux module)


McSinyx:
Thank you for the PR! Since this widget type is refactored, could you also checkout the `async` branch and make a similar change? I could do it myself but I'd like to have someone using the widget to test it.

xaverh:
Sure, I can do that.

PR #87, opened 2019-07-18
BSSID in vicious.widgets.wifiiw

xaverh:
As requested in #85, I tested it and it works as it should.


McSinyx:
Thank you very much for testing it out and adding the feature!

PR #91, opened 2019-08-30
Backlight widget

Danacus:
Hello

I was looking for a blacklight/brightness widget, but I couldn't find one, so I quickly made one myself.
The widget uses the `xbacklight` command to read the brightness level of the display.


contrib/xbacklight_linux.lua
```diff
@@ -0,0 +1,26 @@
...
+local xbacklight = {}
+
+local function xbacklight_cmd(args)
+    local f = io.popen("xbacklight "..args)
+    if f == nil then
+      return nil
+    else
+      local line = f:read("*all")
+      f:close()
+      return line
```
McSinyx: Please avoid `io.popen`. In this case, you can use our local library `spawn.easy_async` which wraps Awesome `awful.spawn`. Please look for other widget types that uses this function for examples of how this can be done.

contrib/xbacklight_linux.lua
```diff
@@ -0,0 +1,26 @@
...
+    end
+end
+
+local function worker(format, warg)
+    if not warg then warg = {} end
+    if type(warg) ~= "table" then warg = { warg } end
+    return { math.floor(tonumber(xbacklight_cmd(table.concat(warg, " ")))) }
```
McSinyx: `type` and `math.floor` should also be localized like `tonumber` and `io.popen` (the later should be deprecated though)

Danacus:
Thanks for letting me know how I can improve the widget. I've only recently started using awesome and I still have to figure some things out. I will apply the proposed modifications today.

Danacus:
I think I have applied all changes that you mentioned, let me know if I missed something.

widgets/xbacklight_linux.lua
```diff
@@ -0,0 +1,31 @@
...
+    if type(warg) ~= "table" then warg = { warg } end
+    spawn.easy_async("xbacklight " .. table.concat(warg, " "),
+        function (stdout) 
+            callback({ math.floor(tonumber(stdout)) }) 
+        end)
+end
+--- }}
+
+return helpers.setasyncall(xbacklight_linux)
+
```
McSinyx: I think this is a trailing newline.

widgets/xbacklight_linux.lua
```diff
@@ -0,0 +1,31 @@
+---------------------------------------------------
+-- Licensed under the GNU General Public License v2
+--  * (c) 2019, Daan V. <daan@dirkvanoverloop.be>
+---------------------------------------------------
+
+-- {{{ Grab environment
+local tonumber = tonumber
+local math = { floor = math.floor }
+local type = type
+local setmetatable = setmetatable
```
McSinyx: `setmetatable` is no longer used--using `luacheck` can help you avoid this

Danacus: I didn't know about luacheck, that's quite a useful tool.

Danacus:
I've written some documentation and did the style changes.

I was thinking about maybe having a string with the display as a argument instead of a table with all arguments that will be passed to xbacklight, since the only argument one might use to get the brightness level is the `--display` option (from what I can tell). However, I have not managed to get the `--display` option to work myself, xrandr returns an error.

One more minor detail I have noticed: shouldn't `setasyncall` be `setasynccall`? Because it appears to be a helper function for an async call and not an async all.

Danacus:
The widget type does work for me, xbacklight works just fine without arguments. I wanted to add an option to the widget type to choose a specific display, in case someone needs that. However, I could not get this option to work. I could also remove the arguments, but like you mentioned, if the widget type is becoming so simple, I might as well just define a custom widget in my rc.lua.

I wasn't really aware of the fact that vicious could also be used as a tool that makes creating your own widgets easier. Perhaps some extra documentation on how to define custom widget types would be nice. I got a bit confused when looking at examples when figuring out how to make a widget type, because of some things being deprecated and new helper functions being introduced.

I guess I can just close this PR, if you think there's no need to include this widget type in the repository.
Your proposed solution is a bit simpler as well.

Thanks for helping me create this widget type nonetheless. I now have a better understanding of how vicious works and will definitely keep using it when I decide to create more custom widget types.

Danacus:
> I wasn't really aware of the fact that vicious could also be used as a tool that makes creating your own widgets easier.

It turns out that's exactly what's written in the first paragraph of the README. Maybe I should learn how to read first. :)

McSinyx:
I'm glad to know that we found a solution for this problem. If you have any idea in the future, you're more than welcomed to file an issue or a PR.

> Perhaps some extra documentation on how to define custom widget types would be nice. I got a bit confused when looking at examples when figuring out how to make a widget type, because of some things being deprecated and new helper functions being introduced.

Definitely! The asynchronous API have been introduced for rather a while now and there is still no proper documentation for it. I will document it ASAP: quite some Awesome upstream issue are actually Vicious fault. Again, thank you for this valuable discussion.

PR #92, opened 2019-10-04
Add a notmuch widget

kiike:
Hello everyone! I was using a Notmuch (a mail indexing and threading program, like mu4e) with a widget that I found lying around somewhere. I realised my widget was disabled for a while, so here's a small widget I just created using the async features of Vicious and Awesome WM. Here's how I use it my config:

```lua
-- Mail widget
local widget_mail = wibox.widget.textbox()
local notmuch_exists = awful.util.file_readable("/usr/bin/notmuch") or
		       awful.util.file_readable("/usr/local/bin/notmuch")

if notmuch_exists then
    vicious.register(widget_mail, vicious.widgets.notmuch,
        function (_, args)
            if args["count"] > 0 then
	       return typicons.render("mail")
	    else
	       return ""
	    end
        end, 60, 'tag:inbox AND NOT tag:killed AND tag:unread AND NOT folder:trash')
end
```
And here's the result in my wibox:

It's really simple, but does the job just fine. Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,7 +399,9 @@ Provides a message count according to an arbitrary Notmuch query.
 
 Supported platforms: platform independent.
 
-Argument: the query that is passed to Notmuch. For instance: `tag:inbox AND tag:unread` returns the number of unread messages with tag "inbox". 
+Argument: the query that is passed to Notmuch. For instance:
+`tag:inbox AND tag:unread` returns the number of unread messages with
+tag "inbox".
 
 Returns a table with string keys containing:
 * `${count}`: the count of messages that match the query
--- a/README.md
+++ b/README.md
@@ -393,6 +393,18 @@ Supported platforms: GNU/Linux, FreeBSD.
     `${down_b}`, `${up_b}`, `${down_kb}`, `${up_kb}`, `${down_mb}`, `${up_mb}`,
     `${down_gb}`, `${up_gb}`.
 
+### vicious.widgets.notmuch
+
+Provides a message count according to an arbitrary Notmuch query.
+
+Supported platforms: platform independent.
+
+Argument: the query that is passed to Notmuch. For instance: `tag:inbox AND tag:unread` returns the number of unread messages with tag "inbox". 
+
+Returns a table with string keys containing:
+* `${count}`: the count of messages that match the query
+
+
 ### vicious.widgets.org
 
 Provides agenda statistics for Emacs org-mode.
--- a/widgets/notmuch_all.lua
+++ b/widgets/notmuch_all.lua
@@ -24,7 +24,7 @@ local spawn = require("vicious.spawn")
 local notmuch = {}
 
 local function parse(stdout, stderr, exitreason, exitcode)
-    local output = {count = "N/A"}
+    local output = { count = "N/A" }
     if exitcode == 0 then output.count = tonumber(stdout) end
     return output
 end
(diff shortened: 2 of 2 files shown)


widgets/notmuch_all.lua
```diff
@@ -0,0 +1,42 @@
...
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU Affero General Public License for more details.
+--
+-- You should have received a copy of the GNU Affero General Public License
+-- along with Vicious.  If not, see <https://www.gnu.org/licenses/>.
+
+local tonumber = tonumber
+local helpers = require "vicious.helpers"
+local spawn = require "vicious.spawn"
```
McSinyx: The `foo"bar"` is actually shorthand for `foo("bar")`, so if you decide to use this style, please remove the space between.

README.md
```diff
@@ -393,6 +393,18 @@ Supported platforms: GNU/Linux, FreeBSD.
     `${down_b}`, `${up_b}`, `${down_kb}`, `${up_kb}`, `${down_mb}`, `${up_mb}`,
     `${down_gb}`, `${up_gb}`.
 
+### vicious.widgets.notmuch
+
+Provides a message count according to an arbitrary Notmuch query.
+
+Supported platforms: platform independent.
+
+Argument: the query that is passed to Notmuch. For instance: `tag:inbox AND tag:unread` returns the number of unread messages with tag "inbox".
```
McSinyx: I notice a trailing space at the end of this line and it's also bit long(er than the 80 column limit).

widgets/notmuch_all.lua
```diff
@@ -0,0 +1,42 @@
...
+local notmuch = {}
+local output = {}
+
+local function parse(stdout, stderr, exitreason, exitcode)
+   if exitcode == 0 then
+      output.count = tonumber(stdout)
+   else
+      output.count = "N/A"
+   end
+   return output
```
McSinyx: Since `output` is solely processed within this function, please declare it within the scope, i.e.

```lua
local output = { count = "N/A" }
if exitcode == 0 then output.count = tonumber(stdout) end
return output
```

kiike: Done! Thanks!

README.md
kiike: I've reformatted that paragraph. Cheers.

widgets/notmuch_all.lua
kiike: It looks better with the parentheses, but I'll follow the style guide you all suggest in the future.

Mic92:
I would in general put spaces after open and before close parentheses `{ 'foo', 'bar', 'baz' }`.

mutlusun:
Though it needs the most space, I prefer the first line in all three examples.

McSinyx:
Since 3 out of 4 voted for extra spacing, please edit the table literals accordingly @kiike.

widgets/notmuch_all.lua
```diff
@@ -24,15 +24,15 @@ local spawn = require("vicious.spawn")
 local notmuch = {}
 
 local function parse(stdout, stderr, exitreason, exitcode)
-    local output = {count = "N/A"}
+    local output = { count = "N/A" }
```
kiike: Space added!

widgets/notmuch_all.lua
```diff
@@ -24,15 +24,15 @@ local spawn = require("vicious.spawn")
...
     if exitcode == 0 then output.count = tonumber(stdout) end
     return output
 end
 
 function notmuch.async(format, warg, callback)
     local cmd = ("notmuch count '^%s'"):format(warg)
 
-    spawn.easy_async(cmd, function(...) callback(parse(...)) end)
+    spawn.easy_async(cmd, function (...) callback(parse(...)) end)
```
kiike: Here too!

widgets/notmuch_all.lua
kiike: @McSinyx  ^^:)

PR #93, opened 2019-10-04
Support different sysctl output formats

kiike:
In the io.popen deprecation PR @McSinyx mentioned that different platforms use different ways of outputting the sysctl info. For instance, in OpenBSD we use "=" between key and value, FreeBSD uses ": " and GNU/Linux uses " = ". Well, I think the sysctl helper should support that and account the difference between systems.

Here's a PR that intends to add such support. It's pretty straightforward and Works For Me(TM) BUT it should be tested in other systems. I've no Linux or FreeBSD systems with Awesome WM around here so please try this PR and let me know if it works for you.

This snippet in `rc.lua` was used to test it:
```lua
helpers = require("vicious.helpers")
helpers.sysctl_async({ "hw.usermem",
		       "hw.physmem"},
		     function(ret)
			for k, v in pairs(ret)
			do print(k, v) end
		     end
)
```
Yielding this output:

Cheers,
Enric

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -245,17 +245,28 @@ function helpers.sysctl_async(path_table, parse)
     spawn.with_line_callback("sysctl " .. path, {
         stdout = function(line)
 	    local separators = {
-	       openbsd = "=",
-	       linux = " = ",
-	       freebsd = ": "
+		freebsd = ": ",
+		linux = " = ",
+		openbsd = "="
 	    }
+            local pattern = ("(.+)%s(.+)"):format(separators[helpers.getos()])
+            local key, value = string.match(line, pattern)
+            ret[key] = value
+	end,
+        stderr = function(line)
+            local messages = {
+                openbsd = {"level name .+ in (.+) is invalid"},
+                linux = {"cannot stat /proc/sys/(.+):",
+                         "permission denied on key '(.+)'"},
+                freebsd = {"unknown oid '(.+)'"}
+            }
 
-	    local pattern = "(.+)%s(.+)"
-	    pattern = pattern:format(separators[helpers.getos()])
-
-            if not string.find(line, "sysctl: unknown oid") then
-                local key, value = string.match(line, pattern)
-                ret[key] = value
+            for i=1,#messages[helpers.getos()] do
+                local key = line:match(messages[helpers.getos()][i])
+                if key then
+                    key = key:gsub("/", ".")
+                    ret[key] = "N/A"
+                end
             end
         end,
         output_done = function() parse(ret) end
--- a/helpers.lua
+++ b/helpers.lua
@@ -244,8 +244,17 @@ function helpers.sysctl_async(path_table, parse)
 
     spawn.with_line_callback("sysctl " .. path, {
         stdout = function(line)
+	    local separators = {
+	       openbsd = "=",
+	       linux = " = ",
+	       freebsd = ": "
+	    }
+
+	    local pattern = "(.+)%s(.+)"
+	    pattern = pattern:format(separators[helpers.getos()])
+
             if not string.find(line, "sysctl: unknown oid") then
-                local key, value = string.match(line, "(.+): (.+)")
+                local key, value = string.match(line, pattern)
                 ret[key] = value
             end
         end,
--- a/helpers.lua
+++ b/helpers.lua
@@ -1,5 +1,6 @@
 ---------------------------------------------------
 -- Licensed under the GNU General Public License v2
+--  * (c) 2019, Enric Morales <me@enric.me>
 --  * (c) 2010, Adrian C. <anrxc@sysphere.org>
 --  * (c) 2009, Rémy C. <shikamaru@mandriva.org>
 --  * (c) 2009, Benedikt Sauer <filmor@gmail.com>
--- a/helpers.lua
+++ b/helpers.lua
@@ -245,32 +245,32 @@ function helpers.sysctl_async(path_table, parse)
 
     spawn.with_line_callback("sysctl " .. path, {
         stdout = function(line)
-	    local separators = {
-		freebsd = ": ",
-		linux = " = ",
-		openbsd = "="
-	    }
+            local separators = {
+                freebsd = ": ",
+                linux = " = ",
+                openbsd = "="
+            }
             local pattern = ("(.+)%s(.+)"):format(separators[helpers.getos()])
             local key, value = string.match(line, pattern)
             ret[key] = value
-	end,
+        end,
         stderr = function(line)
             local messages = {
-                openbsd = {"level name .+ in (.+) is invalid"},
-                linux = {"cannot stat /proc/sys/(.+):",
-                         "permission denied on key '(.+)'"},
-                freebsd = {"unknown oid '(.+)'"}
+                openbsd = { "level name .+ in (.+) is invalid" },
+                linux = { "cannot stat /proc/sys/(.+):",
+                          "permission denied on key '(.+)'" },
+                freebsd = { "unknown oid '(.+)'" }
             }
 
-            for i=1,#messages[helpers.getos()] do
-                local key = line:match(messages[helpers.getos()][i])
+            for _, error_message in ipairs(messages[helpers.getos()]) do
+                local key = line:match(error_message)
                 if key then
                     key = key:gsub("/", ".")
                     ret[key] = "N/A"
                 end
             end
         end,
-        output_done = function() parse(ret) end
+        output_done = function () parse(ret) end
     })
 end
 --  }}}
--- a/helpers.lua
+++ b/helpers.lua
@@ -9,6 +9,7 @@
 ---------------------------------------------------
 
 -- {{{ Grab environment
+local ipairs = ipairs
 local pairs = pairs
 local rawget = rawget
 local require = require
@@ -244,7 +245,7 @@ function helpers.sysctl_async(path_table, parse)
     path = table.concat(path, " ")
 
     spawn.with_line_callback("sysctl " .. path, {
-        stdout = function(line)
+        stdout = function (line)
             local separators = {
                 freebsd = ": ",
                 linux = " = ",
@@ -254,7 +255,7 @@ function helpers.sysctl_async(path_table, parse)
             local key, value = string.match(line, pattern)
             ret[key] = value
         end,
-        stderr = function(line)
+        stderr = function (line)
             local messages = {
                 openbsd = { "level name .+ in (.+) is invalid" },
                 linux = { "cannot stat /proc/sys/(.+):",
--- a/widgets/bat_openbsd.lua
+++ b/widgets/bat_openbsd.lua
@@ -18,6 +18,7 @@
 -- along with Vicious.  If not, see <https://www.gnu.org/licenses/>.
 
 -- {{{ Grab environment
+local pairs = pairs
 local tonumber = tonumber
 local table = {
     insert = table.insert
--- a/widgets/bat_openbsd.lua
+++ b/widgets/bat_openbsd.lua
@@ -29,60 +29,61 @@ local helpers = require("vicious.helpers")
 
 local bat_openbsd = {}
 function bat_openbsd.async(format, warg, callback)
-    local battery = {
-        charging_rate = "hw.sensors.acpi%s.power0",
-        last_full_capacity = "hw.sensors.acpi%s.watthour0",
-        remaining_capacity = "hw.sensors.acpi%s.watthour3",
-        design_capacity = "hw.sensors.acpi%s.watthour4",
-        state = "hw.sensors.acpi%s.raw0",
+    local battery_id = warg or "bat0"
+
+    local fields = {
+	charging_rate = ("hw.sensors.acpi%s.power0"):format(battery_id),
+	last_full_capacity = ("hw.sensors.acpi%s.watthour0"):format(battery_id),
+	remaining_capacity = ("hw.sensors.acpi%s.watthour3"):format(battery_id),
+	design_capacity = ("hw.sensors.acpi%s.watthour4"):format(battery_id),
+	state = ("hw.sensors.acpi%s.raw0"):format(battery_id)
     }
 
-    local sysctl_vars = {}
-    for k, v in pairs(battery) do
-        v = v:format(warg or "bat0")
-        battery[k], sysctl_vars[#sysctl_vars+1] = v, v
-    end
+    local sysctl_args = {}
+    for _, v in pairs(fields) do table.insert(sysctl_args, v) end
 
-    helpers.sysctl_async(sysctl_vars, function(ret)
-    for k, v in pairs(battery) do
-        battery[k] = tonumber(ret[v]:match("(.-) "))
-    end
+    local battery = {}
+    helpers.sysctl_async(sysctl_args, function(ret)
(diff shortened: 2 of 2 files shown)


helpers.lua
```diff
@@ -244,8 +244,17 @@ function helpers.sysctl_async(path_table, parse)
 
     spawn.with_line_callback("sysctl " .. path, {
         stdout = function(line)
+	    local separators = {
+	       openbsd = "=",
```
McSinyx: Vicious uses 4 spaces for indentation.  I assume you uses emacs which somehow always mess up the indentation level somewhere around 3 or 4 spaces :stuck_out_tongue:

helpers.lua
```diff
@@ -244,8 +244,17 @@ function helpers.sysctl_async(path_table, parse)
...
+	       linux = " = ",
```
McSinyx: On GNU/Linux, `sysctl` is usually in `/sbin` and requires superuser permission so I'm not sure about the practical use of it for widgets.

helpers.lua
```diff
@@ -244,8 +244,17 @@ function helpers.sysctl_async(path_table, parse)
...
+	       freebsd = ": "
+	    }
+
+	    local pattern = "(.+)%s(.+)"
+	    pattern = pattern:format(separators[helpers.getos()])
```
McSinyx: I personally believe this is unnecessarily verbose--if the pattern fit in the `match` below within 80 columns, please uses it anonymously.

helpers.lua
```diff
@@ -244,8 +244,17 @@ function helpers.sysctl_async(path_table, parse)
...
+
             if not string.find(line, "sysctl: unknown oid") then
```
McSinyx: Since it is not shown in your screenshot, is `sysctl`'s error on OpenBSD has the same format?

kiike:
Hi @mcsinyx! Thanks for having a look at it!
 
>Vicious uses 4 spaces for indentation. I assume you uses emacs which somehow always mess up the indentation level somewhere around 3 or 4 spaces 😛
 
Indeed. Emacs here with lua-mode. I'll have to check the spacing and reformat it. Good catch!
 
> On GNU/Linux, sysctl is usually in /sbin and requires superuser permission so I'm not sure about the practical use of it for widgets.
 
I could use sysctl just fine on my Raspberry Pi although the output is very different than the BSD's. I don't know what use it could have either, but it doesn't hurt to have it I guess.😋
 
> Since it is not shown in your screenshot, is  sysctl 's error on OpenBSD has the same format?
 
To be honest, I didn't account for that case, since I didn't have a clue as to what that error would be caused by. I'm out for the weekend but I'll look at it on Monday.
 
Cheers!

Sat Oct 05 04:17:42 GMT+02:00 2019 Nguyễn Gia Phong <notifications@github.com>:

helpers.lua
McSinyx: > I could use `sysctl` just fine on my Raspberry Pi although the output is very different than the BSD's. I don't know what use it could have either, but it doesn't hurt to have it I guess. :yum:

I obviously forgot that other than desktop computers, GNU/Linux also run on other hardware where permissions might be configured differently.  BTW, the error under GNU/Linux looks like this

    # sysctl foobar
    sysctl: cannot stat /proc/sys/foobar: No such file or directory

Please update the error handling accordingly along with the one on OpenBSD, preferably in the way you handled the separators.

kiike:
Hi @McSinyx,

I don't know how it works on my computer, but both on OpenBSD on my laptop and GNU/Linux on my Raspberry Pi both send errors to stderr. I reworked the parsing functions so that stderr can capture the different variations of errors and push "N/A" if the key is erroring.

I tried this snippet in a test `rc.lua`
```lua
helpers = require("vicious.helpers")
helpers.sysctl_async({"hw.usermem", "foo.bar", "hw.physmem"},
		     function(ret)
			for k, v in pairs(ret)
			do print(k, v) end
		     end)
```
 and I now get this, which looks nicer to me:
```
hw.usermem	8297222144
hw.physmem	8447561728
foo.bar	N/A
```
Instead of N/A, i could also write "ERROR" or the specific error, like "permission denied" or "unknown oid". I think N/A does the job though. On another note, shouldn't the sysctl command be spawned with an LANG=C (or equivalent) environment so that the output is not translated if the user has another language set?

I also addressed the incorrect indenting and the copyright line.

Let me know what you think

helpers.lua
kiike: Addressed in commit. Thanks! :)

helpers.lua
kiike: I've improved the error handling in a commit. Let me know if that's what you were expecting or what i can change.

helpers.lua
kiike: It was. I could've written it differently. Thanks.

helpers.lua
kiike: It doesn't, but that scenario is now taken care of. I could use help testing it on Linux, and FreeBSD though.

McSinyx:
> I don't know how it works on your computer, but both on OpenBSD on my laptop and GNU/Linux on my Raspberry Pi both send errors to stderr. 

It also prints to stderr on my system.  For FreeBSD, we'd need to wait for @mutlusun confirmation.  While on hold, you can try to refactor the battery widget for OpenBSD to use this new facility.

> I reworked the parsing functions so that stderr can capture the different variations of errors and push "N/A" if the key is erroring.

That's great.  Personally I think `N/A` is enough, since it indicates the field is missing.  On permission issues, I'm thinking about documenting it on a page about Vicious helpers which is currently lacking.

> On another note, shouldn't the sysctl command be spawned with an LANG=C (or equivalent) environment so that the output is not translated if the user has another language set?

Definitely! There's the existing `spawn.with_line_callback_with_shell` you can use for this.

mutlusun:
Hello,
thanks all for your work ;) I will test this PR this weekend on FreeBSD ;)

kiike:
> Hello,
> thanks all for your work ;) I will test this PR this weekend on FreeBSD ;)

Thanks in advance for trying it out. Looking forward to your feedback!

kiike:
> ... While on hold, you can try to refactor the battery widget for OpenBSD to use this new facility.

@McSinyx: done! It was a good way to further practice on async code, which has been above my skill level. I also took the chance to make the variables more readable.

kiike:
Thank you again for the feedback. I'll be out for the weekend so ill work on the styling on Monday. On your last question, sysctl outputs the value followed by the units and a description or comment. That rematching captures the value, discarding the unit and the description.
 
Have a great weekend too!

Sat Oct 12 09:11:36 GMT+02:00 2019 Nguyễn Gia Phong <notifications@github.com>:

helpers.lua
```diff
@@ -244,9 +245,29 @@ function helpers.sysctl_async(path_table, parse)
...
         stdout = function(line)
-            if not string.find(line, "sysctl: unknown oid") then
-                local key, value = string.match(line, "(.+): (.+)")
-                ret[key] = value
+	    local separators = {
+		freebsd = ": ",
+		linux = " = ",
+		openbsd = "="
+	    }
+            local pattern = ("(.+)%s(.+)"):format(separators[helpers.getos()])
```
mutlusun: In our `helpers.lua` there are multiple locations where we call `getos()`. Maybe we could save the output once at startup in a global variable and remove this overhead? (`getos()` is not async) Though, I don't think, it's much overhead ...

McSinyx: Actually the output is saved as `kernel_name` so `io.popen` should be called at most once with the module is initialized.  Deprecate the `getos` completely may reduce a bit of overhead (one call and one conditional) but I agree that it's negligible.

kiike:
You reviews are very thourough, @McSinyx. I am learning a lot about this project's coding style and I hope that my code can be soon on par with the project's.

widgets/bat_openbsd.lua
```diff
@@ -19,67 +19,71 @@
...
+				    battery.design_capacity)
+
+	    -- Pass the following arguments to callback function:
+	    --  * battery state symbol (↯, -, !, + or N/A)
+	    --  * remaining capacity (in percent)
+	    --  * remaining time, as reported by the battery
+	    --  * wear level (in percent)
+	    --  * present_rate (in Watts/hour)
+	    return callback({ state, charge, remaining_time,
+			    wear, battery.charging_rate })
```
McSinyx: `wear` should align with `state` from the line above.

widgets/bat_openbsd.lua
```diff
@@ -19,67 +19,71 @@
...
-                  -- current state
-                  local state = STATES[tonumber(bat_info.raw0)]
+    local sysctl_args = {}
+    for _, v in pairs(fields) do table.insert(sysctl_args, v) end
 
-                  -- battery capacity in percent
-                  local percent = tonumber(
-                      bat_info.watthour3 / bat_info.watthour0 * 100)
+    local battery = {}
+    helpers.sysctl_async(sysctl_args, function(ret)
```
McSinyx: Please add the space before the parameter list of the anonymous function like what you did with #92.

widgets/bat_openbsd.lua
```diff
@@ -19,67 +19,71 @@
...
-            { stdout = function (line)
-                  for key, value in line:gmatch(pattern) do
-                      bat_info[key] = value
-                  end
-              end,
-              output_done = function ()
-                  -- current state
-                  local state = STATES[tonumber(bat_info.raw0)]
+    local sysctl_args = {}
+    for _, v in pairs(fields) do table.insert(sysctl_args, v) end
```
McSinyx: I believe you missed localizing `table.insert` here.

widgets/bat_openbsd.lua
```diff
@@ -19,67 +19,71 @@
...
-    async = function (format, warg, callback)
-        local filter = "hw.sensors.acpi" .. (warg or "bat0")
-        local pattern = filter .. ".(%S+)=(%S+)"
-        local bat_info = {}
+    local fields = {
+	charging_rate = ("hw.sensors.acpi%s.power0"):format(battery_id),
+	last_full_capacity = ("hw.sensors.acpi%s.watthour0"):format(battery_id),
+	remaining_capacity = ("hw.sensors.acpi%s.watthour3"):format(battery_id),
+	design_capacity = ("hw.sensors.acpi%s.watthour4"):format(battery_id),
+	state = ("hw.sensors.acpi%s.raw0"):format(battery_id)
```
McSinyx: We have some problems with the indentations here.

widgets/bat_openbsd.lua
```diff
@@ -19,67 +19,71 @@
...
 
-local STATES = { [0] = "↯",         -- not charging
-                 [1] = "-",         -- discharging
-                 [2] = "!",         -- critical
-                 [3] = "+",         -- charging
-                 [4] = "N/A",       -- unknown status
-                 [255] = "N/A" }    -- unimplemented by the driver
+local bat_openbsd = {}
+function bat_openbsd.async(format, warg, callback)
+    local battery_id = warg or "bat0"
```
McSinyx: Just realized that this is much better than overriding `warg` :wink:

helpers.lua
```diff
@@ -244,9 +245,29 @@ function helpers.sysctl_async(path_table, parse)
...
+            local key, value = string.match(line, pattern)
+            ret[key] = value
+	end,
+        stderr = function(line)
+            local messages = {
+                openbsd = {"level name .+ in (.+) is invalid"},
+                linux = {"cannot stat /proc/sys/(.+):",
+                         "permission denied on key '(.+)'"},
+                freebsd = {"unknown oid '(.+)'"}
```
McSinyx: Please insert the spaces inside the braces.

helpers.lua
```diff
@@ -244,9 +245,29 @@ function helpers.sysctl_async(path_table, parse)
...
+            }
+
+            for i=1,#messages[helpers.getos()] do
+                local key = line:match(messages[helpers.getos()][i])
```
McSinyx: I find this a bit difficult to read.  Maybe `ipairs` could be a better idea?

widgets/bat_openbsd.lua
kiike: Awesome! :)

widgets/bat_openbsd.lua
kiike: done!

kiike:
I wanted to test the changes in my main X11 session (i normally test then in Xephyr) so the emacs changes around whitespace and indent-mode were lost. Silly me! Well, one of my previous commits had mixed tabs and spaces. That should be taken care of now. I also addressed your other comments. I hope I haven't missed any.

widgets/bat_openbsd.lua
kiike: done!

McSinyx:
Answering myself:

> On GNU/Linux, sysctl is usually in /sbin and requires superuser permission
> so I'm not sure about the practical use of it for widgets.

Other widget types simply add /sbin to `PATH`:

https://github.com/vicious-widgets/vicious/blob/a542c5dbe20295eec47c4e75228bd30c3c13ff4b/widgets/wifi_linux.lua#L48

However, it's very likely that normal user doesn't have permission to read /proc/sys.

helpers.lua
```diff
@@ -245,32 +245,32 @@ function helpers.sysctl_async(path_table, parse)
...
-                freebsd = {"unknown oid '(.+)'"}
+                openbsd = { "level name .+ in (.+) is invalid" },
+                linux = { "cannot stat /proc/sys/(.+):",
+                          "permission denied on key '(.+)'" },
+                freebsd = { "unknown oid '(.+)'" }
             }
 
-            for i=1,#messages[helpers.getos()] do
-                local key = line:match(messages[helpers.getos()][i])
+            for _, error_message in ipairs(messages[helpers.getos()]) do
```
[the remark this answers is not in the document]
kiike: done! i am the one who should be sorry for not having checked it more thoroughly. This is the kind of stuff that no matter how many times I look at it, it always looks fine to me 😿

helpers.lua
```diff
@@ -245,32 +245,32 @@ function helpers.sysctl_async(path_table, parse)
...
+                local key = line:match(error_message)
                 if key then
                     key = key:gsub("/", ".")
                     ret[key] = "N/A"
                 end
             end
         end,
-        output_done = function() parse(ret) end
+        output_done = function () parse(ret) end
```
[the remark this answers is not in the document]
kiike: Done!

helpers.lua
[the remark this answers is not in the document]
McSinyx: That's a reason we human work together: improve each other's work!  BTW I've merged the PR, and please consider my proposal (you joining us as a maintainer).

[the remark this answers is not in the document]
kiike: I'll ask @Mic92 right now! In the meantime, I am adapting `widgets.cpu_freebsd` to work on OpenBSD.

McSinyx:
That's right.  I've just tested on my Fedora workstation and `sysctl` is even in user's default PATH (somehow /usr/sbin was decided to be added).  It's still able to display quite many options.

kiike:
@McSinyx IIRC most major distros merged all the */bin directories (/usr/local/bin, /usr/bin, etc/) into one, and Fedora was one of the first ones.

McSinyx:
I'm surprised that they merged sbin too.  Thanks for the info btw.

PR #94, opened 2019-10-16
Add copyright notices and contributing guidelines

### LINKED ISSUE #90: Correct copyright notice
Many widget types have outdated information about copyright authors and time. The header format also doesn't comply with GNU GPL requirements.

To fix this, we can use the commit info from the blames.

McSinyx:
I've just finished my draft for copyright headers and contributing guidelines.  Please review the guides and give your opinions on the improvements could be made.

Merging this would close #90.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,7 +78,7 @@ In any case, thank you very much for your contributions!
 This section introduces a guideline for writing idiomatic, robust
 and future-proof widget type code.
 
-### Whitespces in Expressions and Statements
+### Whitespace in Expressions and Statements
 
 Avoid extraneous whitespace in the following situations:
 
@@ -92,7 +92,7 @@ Avoid extraneous whitespace in the following situations:
 
 * Immediately before a comma, semicolon, or colon.
 * Immediately before the open parenthesis, braces, quote, etc.
-  that starts the argument list of a function call; or the open braket
+  that starts the argument list of a function call; or the open bracket
   that starts an indexing.  In other words, prefer these:
 
   ```lua
@@ -133,7 +133,7 @@ long_function_call(
     baz)
 ```
 
-The closing brace or braket on multiline constructs may either line up under
+The closing brace or bracket on multi-line constructs may either line up under
 the first character of the line that starts the construct, as in:
 
 ```lua
@@ -152,12 +152,14 @@ Trailing right parentheses, however, are not allowed.
 
 ### Maximum Line Length
 
-Limit all *code* lines to a maximum of 80 characters.
-Comments and long strings need not to follow this restriction.
+If possible, try to limit all *code* lines to a maximum
+of 80 characters.  In case you find some lines in your patch would be
+more readable exceeding this limit, feel free to discuss with us.
+Comments and long strings need not to follow this restriction however.
 
-As one might have noticed, the syntatic sugars `f{<fields>}`
+As one might have noticed, the syntactic sugars `f{<fields>}`
 (for `f({<fields>})`) and `f'<string>'` (or `f"<string>"`/`f[[<string>]]`,
-for `f('<string>')`) are espescially prefered to squeeze
+for `f('<string>')`) are especially preferred to squeeze
 the line length to this limit.
 
 ### Blank Lines
@@ -187,9 +189,9 @@ For example,
 local type = type
 local table = { concat = table.concat, insert = table.insert }
 
-local awful = require"awful"
+local awful = require("awful")
 
-local helpers = require"vicious.helpers"
+local helpers = require("vicious.helpers")
 ```
 
 ### String Quotes
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,316 @@
+# How to contribute to Vicious
+
+## Filing an Issue
+
+* Ensure the bug was not already reported by searching GitHub Issues.
+* If you're unable to find an open issue addressing the problem,
+  open a new one.  Be sure to include a title and clear description,
+  as much relevant information as possible, such as Awesome errors,
+  and a config sample or an executable test case
+  (using Vicious as a stand-alone library)
+  demonstrating the expected behavior that is not occurring.
+
+Please re-read your issue once again to avoid a couple of common mistakes
+(you can and should use this as a checklist):
+
+* Is the description of the issue itself sufficient?
+  Make sure that it's obvious
+    - What the problem is
+    - How it could be fixed
+    - How your proposed solution would look like
+* Have you provide the versions of Vicious and related software?
+  We would like to how you installed Vicious, which OS you're using,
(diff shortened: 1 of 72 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,316 @@
...
+2. Related third-party libraries
+3. Local libraries
+
+For example,
+
+```lua
+local type = type
+local table = { concat = table.concat, insert = table.insert }
+
+local awful = require"awful"
```
kiike: Is the shorthand version ie. foo"bar" instead of foo("bar") preferred?

CONTRIBUTING.md
```diff
@@ -0,0 +1,316 @@
...
+### Copyright Header
+
+Vicious is released under the GNU GNU General Public License
+version 2 or later and each contributor holds the copyright
+on their contributing.  To make this collective control effective,
+each source file must include a notice of the following format
+denoting the name of all authors
+
+```lua
+-- <one line to give the program's name and a brief idea of what it does.>
```
kiike: What format should this follow?

CONTRIBUTING.md
```diff
@@ -0,0 +1,316 @@
...
+-- Copyright (C) <year>  <name of author> <<email that can be use for contact>>
```
kiike: Absolutely agree! We could also add [SPDX IDs](https://spdx.org/ids) while we're at it.

CONTRIBUTING.md
```diff
@@ -0,0 +1,316 @@
...
+
+However, declare a variable only when you need it, to avoid declaring it
+without an initial value (and therefore you seldom forget to initialize it).
+Moreover, you shorten the scope of the variable, which increases readability. 
+
+### Copyright Header
+
+Vicious is released under the GNU GNU General Public License
+version 2 or later and each contributor holds the copyright
+on their contributing.  To make this collective control effective,
```
kiike: s/contributing/contribution/

CONTRIBUTING.md
McSinyx: I would not say so: both are OK.  The shorthand is especially preferred when we're running out of columns though.

CONTRIBUTING.md
McSinyx: For widget type I think something like *foo widget type* (*for system bar* if it's not system independent) or *widget type providing blah blah blah*.  I had been adding the filenames to my other projects but then I realized a few things:

* The header sample included in the license is meant for single-file programs.  That's why one should give *the program's name*.  For multi-file ones, we already have the *This file is part of foobar*.
* GNU's projects don't actually add filenames to the header, e.g. take a look at [gcc](https://github.com/gcc-mirror/gcc/blob/master/gcc/c/c-lang.c).
* The filename can be space consuming, especially when the description can be verbose for some of our widget types.  Also if it serves any meaning of getting where it at, it should be the relative path to the file within the project, which is even longer.

CONTRIBUTING.md
McSinyx: While I don't object it, I don't think SPDX is necessary anyhow.  Distros like Debian already examine the source very carefully for the license of every single source file.

CONTRIBUTING.md
```diff
@@ -0,0 +1,316 @@
...
+In this case, and this case only, the trailing comma is acceptable
+to avoid diff noises when more values are added,
+but since Vicious often deal with system APIs which rarely ever change,
+it's occasionally helpful to do so.
+
+Trailing right parentheses, however, are not allowed.
+
+### Maximum Line Length
+
+Limit all *code* lines to a maximum of 80 characters.
```
Mic92: I think this can be relaxed a bit. The code base also contains examples exceeding the limit. This is what systemd for example proposes: https://github.com/systemd/systemd/blob/master/docs/CODING_STYLE.md#formatting

McSinyx: I'm not sure about other people's screen, but my 1600x900 is just enough to fit two 80-col splits.  Maybe systemd's situation is different because their devs tends to use IDEs full-screen w/o splitting?  Of course I'll try to rephrase the guide to make it less harsh though.

BTW in this branch, only comments (which often contain long URLs) and long multi-line strings exceed the 80-col limit; but those are not handled by luacheck `max_code_line_length`.

McSinyx:
@mutlusun and @blueyed, I'd love to hear your opinion on this.

blueyed:
@McSinyx 
I'm OK with anything.. ;) - sorry, but this is huge, and I am more or less just a happy user for now.

CONTRIBUTING.md
Mic92: At some point long lines just become less readable and also github won't show long lines fully when reviewing pull requests.

McSinyx:
> I doesn't like `foo"bar"` or `require"vicious.helpers"` that much and would prefer `foo("bar")` and `require("vicious.helpers")`. But I think this is a question of own preferences ;)

I realized that already when during the async refactoring :wink:  That's why the guide takes no position of preference on either, but only notes that the shorthand version can be useful in case we're running out of columns.  Anyway I tweaked a few examples to show that we're not against the `foo("bar")` style.

> Do we need an extra license header in every file?

[GNU GPLv2 *advises* us to so](https://github.com/vicious-widgets/vicious/blob/master/LICENSE#L282)

> To [make it free software], attach the following notices to the program.
> It is safest to attach them to the start of each source file to most effectively
> convey the exclusion of warranty; and **each file should have at least
> the "copyright" line and a pointer to where the full notice is found.**

IMHO it'd be clearer for future contributors to get the licensing information from the top of each file.  However, I believe the copyright information has to be present to make it effective.

If everyone agrees (approval review), I will merge after this.

Mic92:
> > Do we need an extra license header in every file?
> 
> [GNU GPLv2 _advises_ us to so](https://github.com/vicious-widgets/vicious/blob/master/LICENSE#L282) [](/vicious-widgets/vicious/blob/HEAD@{2019-10-23T06:30:41Z}/LICENSE#L282)
> 
> > To [make it free software], attach the following notices to the program.
> > It is safest to attach them to the start of each source file to most effectively
> > convey the exclusion of warranty; and **each file should have at least
> > the "copyright" line and a pointer to where the full notice is found.**
> 

It is possible to keep this license header pretty short so: https://github.com/systemd/systemd/blob/master/src/random-seed/random-seed.c#L1
However this could be discussed separately.

McSinyx:
> It is possible to keep this license header pretty short so: https://github.com/systemd/systemd/blob/master/src/random-seed/random-seed.c#L1

That was also mentioned by @kiike, however when copyright holders are considered, [it doesn't look very right.](https://github.com/systemd/systemd/blob/ad16158c10dfc3258831a9ff2f1a988214f51653/src/core/ima-setup.h)

> However this could be discussed separately.

If so, I'll be merging this by the end of (my) today.  This is a huge patch and although I've been using this branch for quite a while, I'm still uncertain if I'm missing anything.

PR #95, opened 2019-10-16
[cpu_openbsd] Add a CPU usage widget for OpenBSD

kiike:
Hello everyone!

This PR adds a CPU usage widget for OpenBSD. It currently only outputs the usage across all cores, since I can't seem to be able to use the per-core info. As you can see in this video the output is reasonable. The usage is currently calculated from all the values (system, user, nice, interrupt). Let me know if it's alright style-wise. As for functionality, it is pretty barebones but IMHO sufficient.

Cheers,

Enric

### DIFF
--- a/widgets/cpu_openbsd.lua
+++ b/widgets/cpu_openbsd.lua
@@ -30,7 +30,7 @@ local cpu_openbsd = {}
 
 -- Initialize the table that will contain the ticks spent in each subsystem
 -- values: user, nice, system, spin, interrupts, idle
-cpu_openbsd.ticks = { 0, 0, 0, 0, 0, 0 }
+local ticks = { 0, 0, 0, 0, 0, 0 }
 
 function cpu_openbsd.async(format, warg, callback)
     helpers.sysctl_async({ "kern.cp_time" },
@@ -43,7 +43,7 @@ function cpu_openbsd.async(format, warg, callback)
             local period_ticks = {}
             for i = 1, 6 do
                 table.insert(period_ticks,
-                             current_ticks[i] - cpu_openbsd.ticks[i])
+                             current_ticks[i] - ticks[i])
             end
 
             local cpu_total, cpu_busy = 0, 0
@@ -52,7 +52,7 @@ function cpu_openbsd.async(format, warg, callback)
 
             local cpu_usage = math.ceil((cpu_busy / cpu_total) * 100)
 
-            cpu_openbsd.ticks = current_ticks
+            ticks = current_ticks
             return callback({ cpu_usage })
         end)
 end
--- a/widgets/cpu_openbsd.lua
+++ b/widgets/cpu_openbsd.lua
@@ -1,0 +1,60 @@
+-- cpu_openbsd.lua - provide CPU usage on OpenBSD
+-- Copyright (C) 2019  Enric Morales
+--
+-- This file is part of Vicious.
+--
+-- Vicious is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as
+-- published by the Free Software Foundation, either version 2 of the
+-- License, or (at your option) any later version.
+--
+-- Vicious is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU Affero General Public License for more details.
+--
+-- You should have received a copy of the GNU Affero General Public License
+-- along with Vicious.  If not, see <https://www.gnu.org/licenses/>.
+
+local math = { ceil = math.ceil }
+local string = { gmatch = string.gmatch }
+local table = { insert = table.insert }
+local tonumber = tonumber
+
+local helpers = require("vicious.helpers")
+
+
+-- cpu_openbsd: provides both a helper function that allows reading
+-- the CPU usage on OpenBSD systems.
+local cpu_openbsd = {}
+
+-- Initialize the table that will contain the ticks spent in each subsystem
+-- values: user, nice, system, spin, interrupts, idle
+cpu_openbsd.ticks = { 0, 0, 0, 0, 0, 0 }
+
+function cpu_openbsd.async(format, warg, callback)
+    helpers.sysctl_async({ "kern.cp_time" },
+	function (ret)
+	    local current_ticks = {}
+	    for match in string.gmatch(ret["kern.cp_time"], "(%d+)") do
+		table.insert(current_ticks, tonumber(match))
+	    end
+
+	    local period_ticks = {}
+	    for i=1, 6 do
+		table.insert(period_ticks,
+			     current_ticks[i] - cpu_openbsd.ticks[i])
+	    end
+
+	    local cpu_total, cpu_busy = 0, 0
+	    for i = 1, 6 do cpu_total = cpu_total + period_ticks[i] end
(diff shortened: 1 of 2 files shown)


widgets/cpu_openbsd.lua
```diff
@@ -0,0 +1,60 @@
...
+-- it under the terms of the GNU General Public License as
+-- published by the Free Software Foundation, either version 2 of the
+-- License, or (at your option) any later version.
+--
+-- Vicious is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU Affero General Public License for more details.
+--
+-- You should have received a copy of the GNU Affero General Public License
```
McSinyx: Big OOF to me: I copied the header from my other project and it has the wrong license name :grin:  While I fixed all of that in the copyright branch, I missed it while reviewing your notmuch widget type too.

widgets/cpu_openbsd.lua
```diff
@@ -0,0 +1,60 @@
+-- cpu_openbsd.lua - provide CPU usage on OpenBSD
```
McSinyx: IMHO the filename is redundant (I've checked several GNU projects: they don't seem to include the filenames in the header).  Since there's the same problem with notmuch wtype, I suggest opening a separate PR for it.

widgets/cpu_openbsd.lua
```diff
@@ -0,0 +1,60 @@
...
+local helpers = require("vicious.helpers")
+
+
+-- cpu_openbsd: provides both a helper function that allows reading
+-- the CPU usage on OpenBSD systems.
+local cpu_openbsd = {}
+
+-- Initialize the table that will contain the ticks spent in each subsystem
+-- values: user, nice, system, spin, interrupts, idle
+cpu_openbsd.ticks = { 0, 0, 0, 0, 0, 0 }
```
McSinyx: I believe using local `ticks` here would already be enough.

widgets/cpu_openbsd.lua
```diff
@@ -0,0 +1,60 @@
...
+
+function cpu_openbsd.async(format, warg, callback)
+    helpers.sysctl_async({ "kern.cp_time" },
+	function (ret)
```
McSinyx: Notice the mis-indent here.  I believe we need to fix `lua-mode` once and for all (that's the reason I can't switch to 
Emacs completely: it's too difficult to have per language, leave alone per file, config of indentation).  When you're done, please ping me to check out your dotfiles!

widgets/cpu_openbsd.lua
```diff
@@ -0,0 +1,60 @@
...
+	    local current_ticks = {}
+	    for match in string.gmatch(ret["kern.cp_time"], "(%d+)") do
+		table.insert(current_ticks, tonumber(match))
+	    end
+
+	    local period_ticks = {}
+	    for i=1, 6 do
```
McSinyx: Please surround the `=` with spaces.

widgets/cpu_openbsd.lua
```diff
@@ -0,0 +1,60 @@
...
+
+	    local cpu_total, cpu_busy = 0, 0
+	    for i = 1, 6 do cpu_total = cpu_total + period_ticks[i] end
+	    for i = 1, 5 do cpu_busy = cpu_busy + period_ticks[i] end
+
+	    local cpu_usage = math.ceil((cpu_busy / cpu_total) * 100 )
+
+	    cpu_openbsd.ticks = current_ticks
+	    return callback({ cpu_usage })
+        end)
```
McSinyx: Odd indentation here too.  If you have Vim, you can just add the mode line and `ggVG=` to reindent the whole thing.  TBH I love how Emacs indent Lisp and Python for me, but for less popular (among the Emacs community) languages, auto-styling is a mess.

widgets/cpu_openbsd.lua
kiike: Heh! I thought it was a new GNU license d'oh! I guess removing Affero from line 14 and 16 does the trick, doesn't it?

widgets/cpu_openbsd.lua
kiike: I'll remove both file headers. The battery module also has that line. I guess I got it from the `spawn.lua` file and thought I had to add it.

widgets/cpu_openbsd.lua
kiike: I tried using `local ticks` within the `cpu_openbsd.async` function and the state wouldn't be stored 😭 for reuse when running the function again.

widgets/cpu_openbsd.lua
kiike: I am reworking my lua-mode config and will ping you when done.

widgets/cpu_openbsd.lua
kiike: Yep, my Lua-mode config was incorrect. I am now fixing that as we speak.

widgets/cpu_openbsd.lua
McSinyx: I meant local ticks at the module level (closure).  While it might be obvious to you when you write it, future ready might confused on why `ticks` has to be part of the returning table.  TBH it took me a while.

widgets/cpu_openbsd.lua
McSinyx: For examples, please [navigate to the copyright branch](https://github.com/vicious-widgets/vicious/blob/copyright/widgets/bat_openbsd.lua).

widgets/cpu_openbsd.lua
```diff
@@ -0,0 +1,60 @@
...
+            for i = 1, 6 do
+                table.insert(period_ticks,
+                             current_ticks[i] - cpu_openbsd.ticks[i])
+            end
+
+            local cpu_total, cpu_busy = 0, 0
+            for i = 1, 6 do cpu_total = cpu_total + period_ticks[i] end
+            for i = 1, 5 do cpu_busy = cpu_busy + period_ticks[i] end
+
+            local cpu_usage = math.ceil((cpu_busy / cpu_total) * 100 )
```
McSinyx: I think we missed the last space here.

widgets/cpu_openbsd.lua
McSinyx: Thank you for sharing the config!  Regarding the `.dir-local.el`, you could either add it our `.gitignore` it.  Since you're a maintainer now you'd be working with the sources quite often.

McSinyx: Yep that's what I meant, except there was a typo letting you understood otherwise:

> Regarding the .dir-local.el, you could either add it o~~u~~r .gitignore it.

My dyslexia is stronk!

kiike:
> This looks almost fine to me now. After making `ticks` local
@McSinyx, can you check if that's what you meant please?

>feel free to merge any time! And please remember to squash these fix-ups into the first commit.
As in "combining all these commits" so that they look like the came from 
44eae5e right?

McSinyx:
Yes, it looks perfect to me now!  If you use the Github interface there is an **Squash and merge** option in the drop-down of the green button.  Using git I guess we can do a fix-up rebase and then rebase the newly squashed commit but I've never done it before.

kiike:
OK @McSinyx! Done! I will practice rebasing with a local repo later and push the Emacs dir local later too.
Have a good weekend!

McSinyx:
Thank you, you too!  Although I have classes full Saturday lmfao (the last few days was off however).

PR #97, opened 2020-05-17
Fix: attempt to perform aithmetic on nil value (field 'sub')

elmeriniemela:
Fix pkg widget for Arch, Arch C, FreeBSD and Mandriva


elmeriniemela:
@McSinyx Added my name to copyright header and updated the change log.

McSinyx:
I've just merged this, and thanks again for the catch!

PR #99, opened 2020-08-26
[hwmontemp] Bring back sysfs path cache

lynix:
Cache sysfs path resolved from given name index in local table. This avoids
spawning a `grep` process over and over again for each invocation.

### DIFF
--- a/widgets/hwmontemp_linux.lua
+++ b/widgets/hwmontemp_linux.lua
@@ -1,5 +1,5 @@
 -- widget type providing name-indexed temperatures from /sys/class/hwmon
--- Copyright (C) 2019  Alexander Koch <lynix47@gmail.com>
+-- Copyright (C) 2019, 2020  Alexander Koch <lynix47@gmail.com>
 -- Copyright (C) 2019  Nguyễn Gia Phong <vn.mcsinyx@gmail.com>
 --
 -- This file is part of Vicious.
@@ -25,6 +25,14 @@ local io = { open = io.open }
 local helpers = require"vicious.helpers"
 local spawn = require"vicious.spawn"
 
+local pathcache = {}
+
+local function read_sensor(path, callback)
+    local f = io.open(path, "r")
+    callback{ tonumber(f:read"*line") / 1000 }
+    f:close()
+end
+
 -- hwmontemp: provides name-indexed temps from /sys/class/hwmon
 -- vicious.widgets.hwmontemp
 return helpers.setasyncall{
@@ -32,6 +40,7 @@ return helpers.setasyncall{
         if type(warg) ~= "table" or type(warg[1]) ~= "string" then
             return callback{}
         end
+
         local input = warg[2]
         if type(input) == "number" then
             input = ("temp%d_input"):format(input)
@@ -39,16 +48,20 @@ return helpers.setasyncall{
             input = "temp1_input"
         end
 
-        spawn.easy_async_with_shell(
-            "grep " .. warg[1] .. " -wl /sys/class/hwmon/*/name",
-            function (stdout, stderr, exitreason, exitcode)
-                if exitreason == "exit" and exitcode == 0 then
-                    local f = io.open(stdout:gsub("name%s+", input), "r")
-                    callback{ tonumber(f:read"*line") / 1000 }
-                    f:close()
-                else
-                    callback{}
-                end
-            end)
+        local sensor = warg[1]
+        if pathcache[sensor] then
+            read_sensor(pathcache[sensor] .. input, callback)
+        else
+            spawn.easy_async_with_shell(
+                "grep " .. sensor .. " -wl /sys/class/hwmon/*/name",
+                function (stdout, stderr, exitreason, exitcode)
+                    if exitreason == "exit" and exitcode == 0 then
+                        pathcache[sensor] = stdout:gsub("name%s+", "")
+                        read_sensor(pathcache[sensor] .. input, callback)
+                    else
+                        callback{}
+                    end
+                end)
+        end
     end }
 -- vim: ts=4:sw=4:expandtab


widgets/hwmontemp_linux.lua
```diff
@@ -39,16 +42,21 @@ return helpers.setasyncall{
...
+        if not pathcache[warg[1]] then
+            spawn.easy_async_with_shell(
+                "grep " .. warg[1] .. " -wl /sys/class/hwmon/*/name",
+                function (stdout, stderr, exitreason, exitcode)
+                    if exitreason == "exit" and exitcode == 0 then
+                        pathcache[warg[1]] = stdout:gsub("name%s+", "")
+                    end
+                end)
+        end
+        if pathcache[warg[1]] then
```
McSinyx, line 54: On the first run this is very likely to be reached before the `pathcache[warg[1]]` is assigned (also nit pick: please give `warg[1]` a name, it's used in quite a few places in this patch).  I imagine that you can call `callback` within `easy_async_with_shell` and in the `else`-clause of the block above.  You might need to define a function to avoid duplicating the file opening code though.

lynix: Okay I seem to have misunderstood how this async stuff works, thanks for pointing out.

Well, as the entire worker function (to my understanding) is already asynchronous, would it be an option to replace the `spawn.easy_async_with_shell()` part with something synchronous that performs the `grep` call?

Regarding `warg[1]` you're absolutely right and it was pretty hard for me not to give it a name! But I had a look at other widgets and, together with _"However, declare a variable only when you need it"_ (from CONTRIBUTING.md), I got the impression that one shall really stick to the absolute minimum.

I'll come up with a new pull request including a name for `warg[1]` once we're done talking about the point above.

McSinyx: The worker function (I assume you're referring to the `function (format, warg, callback)` one) is supposed to be asynchronous, which is made so by using `spawn.easy_async_with_shell()` while doing the heavy lifting.  When grep is not needed, I suppose that we can execute the code (which is just reading a tiny file) synchronously w/o risking making the WM unresponsive.

> However, declare a variable only when you need it

IIRC refers to not doing the variable declaration at the beginning of every function like in ancient C (after reading the surrounding sentences that's what I felt it meant).  For the name suggestion, according to [the docs you wrote](https://github.com/vicious-widgets/vicious#viciouswidgetshwmontemp), I think we can call it `sensor`.

widgets/hwmontemp_linux.lua
```diff
@@ -1,6 +1,7 @@
 -- widget type providing name-indexed temperatures from /sys/class/hwmon
 -- Copyright (C) 2019  Alexander Koch <lynix47@gmail.com>
 -- Copyright (C) 2019  Nguyễn Gia Phong <vn.mcsinyx@gmail.com>
+-- Copyright (C) 2020  Alexander Koch <lynix47@gmail.com>
```
McSinyx, line 4:
```suggestion
-- Copyright (C) 2019, 2020  Alexander Koch <lynix47@gmail.com>
-- Copyright (C) 2019  Nguyễn Gia Phong <vn.mcsinyx@gmail.com>
```

widgets/hwmontemp_linux.lua
lynix: > Please don't, just push changes to this one directly
Shame on me, I didn't know you can actually work on pull requests and change things. I'll find out how to do so :)

widgets/hwmontemp_linux.lua
```diff
@@ -25,30 +25,45 @@ local io = { open = io.open }
...
 local spawn = require"vicious.spawn"
 
+local pathcache = {}
+
+local function read_sensor(path)
+    local f = io.open(path, "r")
+    local v = tonumber(f:read"*line") / 1000
+    f:close()
+
+    return v
```
McSinyx, line 35: I think passing callback here would allow us to avoid, ehm, this.

widgets/hwmontemp_linux.lua
```diff
@@ -25,30 +25,45 @@ local io = { open = io.open }
...
+            spawn.easy_async_with_shell(
+                "grep " .. sensor .. " -wl /sys/class/hwmon/*/name",
+                function (stdout, stderr, exitreason, exitcode)
+                    if exitreason == "exit" and exitcode == 0 then
+                        pathcache[sensor] = stdout:gsub("name%s+", "")
+                        callback{ read_sensor(pathcache[sensor] .. input) }
+                    else
+                        callback{}
+                    end
+                end)
```
McSinyx, line 66: Something like the following will not rerun grep if the given sensor is not found:

```suggestion
        if pathcache[sensor] == nil then
            spawn.easy_async_with_shell(
                "grep " .. sensor .. " -wl /sys/class/hwmon/*/name",
                function (stdout, stderr, exitreason, exitcode)
                    if exitreason == "exit" and exitcode == 0 then
                        pathcache[sensor] = stdout:gsub("name%s+", "")
                        callback{ read_sensor(pathcache[sensor] .. input) }
                    else
                        pathcache[sensor] = false
                        callback{}
                    end
                end)
        elseif pathcache[sensor] then
            callback{ read_sensor(pathcache[sensor] .. input) }
        else
            callback{}
```

McSinyx (changes requested):

widgets/hwmontemp_linux.lua
lynix: Re-running `grep` if the sensor could not be found was actually intended. I thought of the (admittedly rare) case of hot-pluggable devices that may occur after the window manager has been started.

I think it's a matter of taste. Invalid sensors should not be the average case but I get your point.

widgets/hwmontemp_linux.lua
lynix: I'm not entirely sure if I understand what you mean. Getting rid of `local v` and calling the passed callback inside of `read_sensor()`?

So that would save us the variable at the cost of an additional parameter that has to be passed. Well, I guess I'm too much of a C guy, trusting the compiler to optimize at this level ;)

widgets/hwmontemp_linux.lua
McSinyx: Sounds fair to me.

widgets/hwmontemp_linux.lua
McSinyx: It was not exactly about (micro-)optimization, I just don't like the *let me save a value as some variable and do something and return it later* thing.  I don't feel really strong about this though so if I can't change your mind, feel free to resolve the conversation and I'll merge.

lynix: Okay I understand. I'll come up with a code proposal and force-push this pull request branch.

It's going to be at least tomorrow as I'll have no internet access for the rest of today.

McSinyx (approved):

Issue #102, opened 2020-09-08
Awesome stalls after calling vicious.call for async widgets


Sajohn-CH:
I noticed that awesome seems to stall if I use 'vicious.call' on some of the widgets (See code below which). Awesome does not react and I get a 100% cpu usage (for one core) used by awesome. Is this a bug or did I do the calls incorrectly?

I used the current master branch (commit f452b57) and awesome version 4.3.4 using Lua 5.3.5 on Fedora using kernel 5.8.4. A minimal working config, which shows the behaviour as soon as one uncomments one of the prints using vicious.call.
I ran it using Xephyr to be able to see the output of the prints.

```
-- If LuaRocks is installed, make sure that packages installed through it are
-- found (e.g. lgi). If LuaRocks is not installed, do nothing.
pcall(require, "luarocks.loader")

-- Standard awesome library
gears = require("gears")
awful = require("awful")
require("awful.autofocus")
-- Widget and layout library
wibox = require("wibox")
-- Theme handling library
beautiful = require("beautiful")
-- Notification library
naughty = require("naughty")
menubar = require("menubar")
hotkeys_popup = require("awful.hotkeys_popup")
-- Enable hotkeys help widget for VIM and other apps
-- when client with a matching name is opened:
require("awful.hotkeys_popup.keys")

-- include vivious widgets
vicious = require("vicious")

-- {{{ Error handling
-- Check if awesome encountered an error during startup and fell back to
-- another config (This code will only ever execute for the fallback config)
if awesome.startup_errors then
    naughty.notify({ preset = naughty.config.presets.critical,
                     title = "Oops, there were errors during startup!",
                     text = awesome.startup_errors })
end

-- Handle runtime errors after startup
do
    local in_error = false
    awesome.connect_signal("debug::error", function (err)
        -- Make sure we don't go into an endless error loop
        if in_error then return end
        in_error = true

        naughty.notify({ preset = naughty.config.presets.critical,
                         title = "Oops, an error happened!",
                         text = tostring(err) })
        in_error = false
    end)
end
-- }}}

-- works
print("battery: "..vicious.call(vicious.widgets.bat, "$3", "0"))
print("cpu: "..vicious.call(vicious.widgets.cpu, "$1"))
print("cpufreq: "..vicious.call(vicious.widgets.cpufreq, "$1", "cpu0"))
print("cpuinf: "..vicious.call(vicious.widgets.cpuinf, "${cpu0 mhz}"))
[...]

PR #101, opened 2020-09-07
Move documentation to Sphinx

McSinyx:
Resolve GH-100, or at least create a concrete point for discussion.

@Mic92, may I have this repo's admin permission to config the web hooks (and other things in the future)?

PR #103, opened 2020-09-14
Separate vicious.call into sync and async versions

McSinyx:
Fix GH-102.  Since OP is happy with the work-around, so I'll wait for GH-101 before writing docs and changelog for this one.

As for the noises at the beginning of the file, I grouped copyright entries that were not recognized as of the same person by `headergen`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,8 +1,16 @@
 Changelog
 =========
 
-Changes in 2.5.0 (WIP)
-----------------------
+Changes in 2.5.0
+----------------
+
+Fixed:
+
+- ``vicious.call`` freezing awesome when used with asynchronous widget types
+
+Added:
+
+- ``vicious.call_async`` asynchronous analogous to ``vicious.call``
 
 Moved:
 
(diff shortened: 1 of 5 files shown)


McSinyx:
Irrelevant change: I switched to Read the Docs default theme to be more mobile-friendly.

Mic92 (approved):

McSinyx:
@Sajohn-CH, could you please try if [vicious.call_async](https://vicious--103.org.readthedocs.build/en/103/usage-awesome.html#vicious.call_async) works better for you than the suggest workaround?

@mutlusun, do you have any comment on this or it's OK for me to go ahead and merge?

Sajohn-CH:
@McSinyx I'll try it on the weekend

Sajohn-CH:
Tank you for implementing a fix. I had a look at it.
For my purpose, the workaround with (vicious.widgets.<widget>.async) and the new vicious.call_async work equally well. I used them to get the values of widgets (e.g. volume) to show in a popup when I pressed the key to change the volume (in awesomewm). Personally I think vicious.call_async is easier to use as it is closer to vicious.register as you can pass a format string or function.

McSinyx:
Thanks for the feedback, I'm gonna go ahead and merge this now, any problem to be found can be knocked out later.

PR #104, opened 2020-09-26
[contrib.openweather] added missing app id and some additional return variables

MArpogaus:
The current implementation is not working due to the lack of an API key (appid).

### Changes
 * Added app_id to warg.
 * New return variables {temp min c}, {temp… max c}, {sunrise} and {sunset}

### New behavior

 * Argument: a table with fields `city_id` (e.g. number: 2643743) and `app_id` (e.g string: '4c57f0c88d9844630327623633ce269cf826ab99')
 * Returns a table with string keys: `${city}`, `${humid}`, `${press}`, `${sky}`, `${sunrise}`, `${sunset}`, `${temp c}`, `${temp max c}`, `${temp min c}`, `${weather}`, `${wind aim}`, `${wind deg}`, `${wind kmh}`, `${wind mps}`

### DIFF
--- a/contrib/openweather_all.lua
+++ b/contrib/openweather_all.lua
@@ -19,8 +19,6 @@
 -- along with Vicious.  If not, see <https://www.gnu.org/licenses/>.
 -- {{{ Grab environment
 local tonumber = tonumber
-local io = {popen = io.popen}
-local setmetatable = setmetatable
 local string = {match = string.match}
 local math = {ceil = math.ceil, floor = math.floor}
 local helpers = require "vicious.helpers"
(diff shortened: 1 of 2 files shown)


McSinyx:
Thanks for the contribution!  Since it's in the contrib collection, Imma be chill on this PR and ignore the styling changes and won't question backward compatibility (I've thought of promoting this to a supported widget type, but having to have an API ID to maintain it doesn't appeal to me :smile:

You might want to add your name to the copyright header and convert this to an asynchronous widget type though.  Either way I have no objection against this patch.

MArpogaus:
I've done the suggested modifications and updated the docs.

McSinyx:
Thanks!

PR #106, opened 2020-10-18
vicious.widgets.gmail docs update

### LINKED ISSUE #105: vicious.widgets.gmail docs update
Hi!
Using [these](https://pythoncircle.com/post/727/accessing-gmail-inbox-using-python-imaplib-module/) instructions I was able to use `vicious.widgets.gmail` without disabling two step verification. Can someone update the docs?

senyai:
more info on getting gmail password. Fixes #105

### DIFF
--- a/docs/source/widgets.rst
+++ b/docs/source/widgets.rst
@@ -150,8 +150,10 @@ Provides count of new and subject of last e-mail on Gmail.
 Supported platform: platform independent, requiring ``curl``.
 
 This widget expects login information in your ``~/.netrc`` file, e.g.
-``machine mail.google.com login user password pass`` and you have to disable
-`two step verification`_.  `Allow access for less secure apps`_ afterwards.
+``machine mail.google.com login user password pass``. If you have 2FA
+account, you can use 'App password' which is created on
+`Google Security page`_. Otherwise, you have to disable
+`two step verification`_ and `allow access for less secure apps`_.
 
 .. caution::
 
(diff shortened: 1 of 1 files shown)


docs/source/widgets.rst
```diff
@@ -501,7 +503,8 @@ Supported platforms: GNU/Linux.
   and ``${sign}`` (signal level, in dBm)
 
 .. _strftime(3): https://linux.die.net/man/3/strftime
+.. _Google Security page: https://myaccount.google.com/security?gar=1
```
McSinyx, line 506: From [the instruction I found], it seems that the app password thing is hidden behind the *Signing in to Google* thing.  Is that still the case?  If so you might want to point to that instruction instead (I also assume that link will be valid even if Google change the procedure in the future).

docs/source/widgets.rst
```diff
@@ -150,8 +150,10 @@ Provides count of new and subject of last e-mail on Gmail.
 Supported platform: platform independent, requiring ``curl``.
 
 This widget expects login information in your ``~/.netrc`` file, e.g.
-``machine mail.google.com login user password pass`` and you have to disable
-`two step verification`_.  `Allow access for less secure apps`_ afterwards.
+``machine mail.google.com login user password pass``. If you have 2FA
+account, you can use 'App password' which is created on
+`Google Security page`_. Otherwise, you have to disable
+`two step verification`_ and `allow access for less secure apps`_.
```
McSinyx, line 156: I think 2FA and two step verification mean the same thing.  Sorry for introducing the confusion in the linked issue :sweat_smile:

McSinyx (changes requested):
Thank you for taking up this!  There are a few spots needing some alteration though:

Issue #109, opened 2022-04-17
MPD progress broken / format not properly escaped


cpiber:
awesome v4.2 (Human after all)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.5
 • LGI version: 0.9.2

`${Progress}` returns a percentage with `%` appended. `helpers.format` however does not properly escape arguments, resulting in invalid patterns.

```diff
diff --git a/helpers.lua b/helpers.lua
index 1e3b198..b24bb7a 100644
--- a/helpers.lua
+++ b/helpers.lua
@@ -164,9 +164,10 @@ end
 -- {{{ Format a string with args
 function helpers.format(format, args)
     for var, val in pairs(args) do
+        local escaped = type(val) == "string" and val:gsub("%%", "%%%%") or val
         format = format:gsub("$" .. (tonumber(var) and var or
             var:gsub("[-+?*]", function(i) return "%"..i end)),
-        val)
+        escaped)
     end
 
     return format
```

Note: in any case this is an inconsistency, mpd `${Progress}` returns data with `%`, but most other widgets will return them as just a number...

McSinyx:
Thank you for reporting!  Would you like to submit a PR?

cpiber:
Yes, I can do that

PR #110, opened 2022-04-19
(#109) Escape % before using strings in gsub

cpiber:
As discussed, here's the PR.


McSinyx:
Thanks, I reformatted it a bit and added you to the copyright header.

PR #111, opened 2022-04-19
Catch errors in update_value to release lock

cpiber:
If an error happens in update_value, reg.lock is never released, thus resulting in a deadlock. This makes sure to catch such errors and merely output a warning.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -38,6 +38,13 @@ local table = {
     remove  = table.remove
 }
 local helpers = require("vicious.helpers")
+local dstatus, debug = pcall(require, "gears.debug")
+local warn
+if dstatus then
+  warn = debug.print_warning
+else
+  warn = function (msg) io.stderr:write('Warning (vicious): ', msg, '\n') end
+end
 
 -- Vicious: widgets for the awesome window manager
 local vicious = {}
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -38,6 +38,13 @@ local table = {
     remove  = table.remove
 }
 local helpers = require("vicious.helpers")
+local dstatus, debug = pcall(require, "gears.debug")
```
McSinyx, line 41: Shouldn't naughty.notify be used here instead?

init.lua
```diff
@@ -38,6 +38,13 @@ local table = {
...
+local warn
+if dstatus then
+  warn = debug.print_warning
+else
+  warn = function (msg) io.stderr:write('Warning (vicious): ', msg, '\n') end
```
McSinyx, line 46: io.stderr should be localized here, also for the consistency with the few lines above, double quote might be preferred.

McSinyx (changes requested):
Could you please provide a scenario where update_value fails for me to test?  The normal case works fine.

In addition, please add your name to the copyright header.

cpiber:
A scenario where it consistently fails for me is when `mpd` is turned off

init.lua
cpiber: I can change it, but wouldn't that spam notifications?

McSinyx: I was genuinely asking :smiley:  It makes sense to log instead then.

McSinyx:
Thanks!

PR #112, opened 2022-04-19
Add ability to send arbitrary commands to MPD

cpiber:
This PR allows the user to send commands to MPD.
I have currently implemented wrappers for some of the more common commands. Maybe it would be a good idea to expose the `build_cmd` to allow users to add their own.

I fully understand if you feel this is not in the scope of this library :)

### DIFF
--- a/widgets/mpd_all.lua
+++ b/widgets/mpd_all.lua
@@ -5,6 +5,7 @@
 -- Copyright (C) 2018-2019  Nguyễn Gia Phong <vn.mcsinyx@gmail.com>
 -- Copyright (C) 2019  Juan Carlos Menonita <JuanKman94@users.noreply.github.com>
 -- Copyright (C) 2019  Lorenzo Gaggini <lg@lgaggini.net>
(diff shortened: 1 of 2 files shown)


widgets/mpd_all.lua
```diff
@@ -83,8 +117,8 @@ function mpd_all.async(format, warg, callback)
         ["{bitrate}"]  = 0,
         ["{elapsed}"]  = 0,
         ["{duration}"] = 0,
-        ["{repeat}"]   = false,
-        ["{random}"]   = false,
+        ["{repeat}"]   = 0,
+        ["{random}"]   = 0,
```
cpiber, line 121: This is a simple change to make sure data types stay the same, since mpd will send `0` or `1`, not `true` or `false`

McSinyx (changes requested):
Sorry it took a while for me to get back to this, could you please update the documentation in `docs/source/widgets.rst` accordingly?  Please add yourself to the `mpd_all.lua` copyright header as well.

McSinyx:
Thanks!

PR #113, opened 2022-06-08
Made module require relative

thomascft:
I changed the require statements to be relative rather than absolute to the config root. This allows for vicious to be put in a folder such as `modules` and accessed with `require("modules.vicious")` without it breaking.  I'm not sure about versioning and such so... I'll leave it up to Y'all. If you want an example of the changes I made in action under AwesomeWM take a look at [Bling](https://github.com/blingcorp/bling) and how their require statements are formatted.

### DIFF
--- a/widgets/init.lua
+++ b/widgets/init.lua
@@ -19,6 +19,6 @@
 -- along with Vicious.  If not, see <https://www.gnu.org/licenses/>.
 
 local setmetatable = setmetatable
-local wrequire = require("vicious.helpers").wrequire
+local wrequire = require((...):match(".*vicious") .. ".helpers").wrequire
 
-return setmetatable({ _NAME = "vicious.widgets" }, { __index = wrequire })
+return setmetatable({ _NAME = (...):match(".*vicious") .. ".widgets" }, { __index = wrequire })
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -201,7 +201,7 @@ For example,
 
    local awful = require("awful")
 
-   local helpers = require("vicious.helpers")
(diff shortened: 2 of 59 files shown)


widgets/init.lua
```diff
@@ -19,6 +19,6 @@
 -- along with Vicious.  If not, see <https://www.gnu.org/licenses/>.
 
 local setmetatable = setmetatable
-local wrequire = require("vicious.helpers").wrequire
+local wrequire = require((...):match(".*vicious") .. ".helpers").wrequire
 
-return setmetatable({ _NAME = "vicious.widgets" }, { __index = wrequire })
+return setmetatable({ _NAME = (...):match(".*vicious") .. ".widgets" }, { __index = wrequire })
```
McSinyx, line 24: This line is too long.

McSinyx:
> This allows for vicious to be put in a [parent package]

Do you actually use this or is this a hypothetical case?  I don't have any opposition against this, but only if it is useful in practice.

> I'm not sure about versioning and such so...

Could you please elaborate a bit?  I have yet to see immediate issue related to versioning this could cause.

thomascft:
I like to have my extra libraries in a folder rather than in config root. It keeps the file structure cleaner. Usually I put them in a folder labeled modules, but it didn't work with this. So this pr makes it so people can do that rather than having extra stuff in the config root.

thomascft:
Oh by versioning I meant the changelog. I'm not sure how that's structured.

widgets/init.lua
thomascft: I guess I can make the path a variable and then assign it.

PR #115, opened 2022-12-25
[amdgpu_linux] Add widget type

shikamaru:
Hi guys,

I’ve had this vicious widget I made about 2 years ago, I use both of the metrics in my config, one with a progressbar and the other with a graph, it works quite well for me. It’s been a long time since I last contributed to vicious (12 years!), hope you guys like this one !

Merry Christmas,

Regards,

### DIFF
--- a/widgets/amdgpu_linux.lua
+++ b/widgets/amdgpu_linux.lua
@@ -1,0 +1,62 @@
+-- contrib/amdgpu_linux.lua
+-- Copyright (C) 2022  Rémy Clouard <shikamaru shikamaru fr>
+--
+-- This file is part of Vicious.
+--
+-- Vicious is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as
+-- published by the Free Software Foundation, either version 2 of the
+-- License, or (at your option) any later version.
+--
+-- Vicious is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with Vicious.  If not, see <https://www.gnu.org/licenses/>.
+local io = { open = io.open }
+local helpers = require("vicious.helpers")
+
+-- {{{ AMDGPU widget type
+return helpers.setcall(function (format, warg)
+    if not warg then return end
+
+    local amdgpu = "/sys/class/drm/"..warg.."/device"
+    local _data = {}
+    local _mem = nil
+
+    local f = io.open(amdgpu .. "/gpu_busy_percent", "r")
+    if f then
+        for line in f:lines() do
+            _data["{gpu_usage}"] = line
+        end
+        f:close()
+    else
+        _data["{gpu_usage}"] = "N/A"
+    end
+
+    if _mem == nil then
+        f = io.open(amdgpu .. "/mem_info_vram_total", "r")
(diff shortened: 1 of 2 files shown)


widgets/amdgpu_linux.lua
```diff
@@ -0,0 +1,62 @@
...
+    if f then
+        for line in f:lines() do
+            _data["{gpu_usage}"] = line
+        end
+        f:close()
+    else
+        _data["{gpu_usage}"] = "N/A"
+    end
+
+    if _mem == nil then
```
McSinyx, line 39: My Lua is rusty now but I suppose `_mem == nil` should always be true here.

McSinyx, line 39: My Lua is rusty now but I suppose `_mem == nil` should always be true here.

widgets/amdgpu_linux.lua
```diff
@@ -0,0 +1,62 @@
...
+                _mem = line
+            end
+            f:close()
+        end
+    end
+
+    f = io.open(amdgpu .. "/mem_info_vram_used", "r")
+    if f then
+        for line in f:lines() do
+            _data["{mem_usage}"] = line/_mem*100
```
McSinyx, line 52: I have a feeling `line` and `_mem` should be checked to be numbers instead of propagating an *attempt to div* error to naughtification.

McSinyx (commented):
Thank you for the Christmas present!  Since I don't have an AMD GPU and it's already working for you, I only left comments on the possible edge cases.  Does AMD provide any specification on the presence and format of sysfs files?

widgets/amdgpu_linux.lua
shikamaru: yeah, I wanted to put that variable in a closure based on the assumption that the max vram wouldn’t change on subsequent calls, but atm it’s a local variable which means it’s called every time, so it could be optimized there, I admit I’m not a lua expert either, how would you do that, define that variable outside of the function ? In practice I don’t notice a negative performance effect but I don’t update the associated widget as often as the performance graph.

shikamaru:
Thanks for the review, I didn’t use everything that the open source amdgpu driver provides, but yeah, a comment pointing to this documentation would probably be helpful : https://www.kernel.org/doc/html/v5.9/gpu/amdgpu.html

widgets/amdgpu_linux.lua
McSinyx: > I wanted to put that variable in a closure based on the assumption that the max vram wouldn’t change on subsequent calls, [...], how would you do that, define that variable outside of the function ?

Yes, some other widgets also do it.

widgets/amdgpu_linux.lua
McSinyx: Looking at the specs, it seems mem_info_vram_total and mem_info_vram_used should exist together, but at least one should be checked with tonumber. They should also consist of only one line, thus f:read'*l' can be preferred for better communicating the intention.

shikamaru:
@McSinyx here is what I came up with, I checked with strace now mem_info_vram_total is only read once, I checked for numbers and prevent division by zero and added a link to the documentation. I missed your comment about using f:read, just modified that as well.

McSinyx (approved):
LGTM.  `_data`'s fields could be initialized as `N/A` for better concision but it's a matter of taste.  I will be merging this around new year's eve, just in case there's something we missed and for other maintainers to chime in their opinions, if any.

McSinyx:
:sweat_smile: I failed my own deadline but happy new year nevertheless :fireworks:

shikamaru:
thank you man, happy new year!

PR #118, opened 2024-08-14
docs: Update usage of batwidget with updated api for wibox margin examples.rst

lumpsoid:
I'm new to awesomewm configuration and it was a problem to understand why battery example widget is not shown on the top bar, after I added the example as is to my config, because awesomewm don't provide any errors when loading config with previous version of example.

but they changed the api for `wibox.layout.margin`, now it is `wibox.container.margin`([link](https://awesomewm.org/doc/api/classes/wibox.container.margin.html))

after changing it in my config, I got this widget on my top bar.


McSinyx (approved):
Great catch, thanks!

Issue #119, opened 2024-10-25
viscious.widgets.bat (dis)charge rate is always N/A


arch-stack:
When I use viscious.widgets.bat on linux, $5 the battery (dis)charge rate is always N/A

I am using a framework 13 laptop with arch linux and it appears whichever battery module is loaded is not providing the power_now file that is used in https://github.com/vicious-widgets/vicious/blob/master/widgets/bat_linux.lua#L49C5-L51C8

While

> /sys/class/power_supply/BAT1/power_now

is missing

> /sys/class/power_supply/BAT1/current_now
> /sys/class/power_supply/BAT1/voltage_now

are present, which means we should be able to use those as a fallback

We could change:

```
if battery.power_now then
    curpower = string.format("%.2f", tonumber(battery.power_now) /1000000)
end
```

To:

```
if battery.power_now then
    curpower = string.format("%.2f", tonumber(battery.power_now) /1000000)
elseif battery.current_now and battery.voltage_now then
    curpower = string.format("%.2f", tonumber(battery.current_now) * tonumber(battery.voltage_now) / 10^12)
end
```

McSinyx:
Thank you for reporting.  Your patch seems reasonable, could you please [open a PR](https://vicious.readthedocs.io/en/latest/contributing.html#requesting-for-merging-a-patch)?

arch-stack:
I just about have a PR ready, the only thing is I only have it tested in lua 5.3 as that appears to be what arch has awesomewm compiled against

Should be fine for such a small change based on existing code?

arch-stack:
I have opened a PR for this issue

So far no issues during my occasional usage over the last week

McSinyx:
Thanks, I've just released v2.7.1 including your patch in case you want to update Vicious for your distro.

PR #120, opened 2024-11-08
[bat_linux] Fallback to computing bat (dis)charge rate in some cases

arch-stack:
This PR is to fix the issue in #119 

Some devices/batteries do not provide the `/sys/class/power_supply/BAT#/power_now` file but they do provide the `/sys/class/power_supply/BAT#/current_now` and `/sys/class/power_supply/BAT#/voltage_now` files

Using the basic P=EI formula (or P=VI if you prefer) to compute power we can take the voltage and current values and multiply them to get the same kind of power value instead of always showing N/A

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+Changes in 2.7.1
+----------------
+
+Fixed:
+
+- [bat_linux] Fallback to computing bat (dis)charge rate in some cases
+
 Changes in 2.7.0
 ----------------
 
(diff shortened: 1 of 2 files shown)


arch-stack:
I would like to mention that I have tested using the default version of awesomewm on arch linux which is compiled against lua 5.3.6

The changes are mostly a copy of the existing functionality so there prob shouldn't be any issues with 5.1 and 5.2 but if someone could lend a hand with those it would be helpful

McSinyx:
Thanks, linter is happy and the patch only affects the previously unhandled path so it should be good!

